PR #2, opened 2017-04-23
Support aliases

AlfonsoUceda:
With this PR you can create several names for a command

Example
--
```ruby
module Foo
  module CLI
    include Hanami::Cli
 
    class Version
      def call
        puts "v1.0.0"
      end
    end

    register 'version', '-v', '--version', Version
  end
end
```
```shell
$ foo version
v1.0.0
$ foo -v
v1.0.0
$ foo --version
v1.0.0
```

I use this version for now, there is another alternative:
```ruby
register ['version', '-v', '--version'], Version
```
But I prefer the first one because we don't have to remember to write the bracket.

cc @jodosha @oana-sipos

### DIFF
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -26,8 +26,8 @@ module Hanami
         true
       end
 
-      def register(name, command)
-        Hanami::Cli.register(name, command)
+      def register(*names, command)
+        names.each { |name| Hanami::Cli.register(name, command) }
       end
     end
 
--- a/spec/support/fixtures/foo
+++ b/spec/support/fixtures/foo
@@ -18,6 +18,12 @@ module Foo
     end
 
     module Generate
+      class Help
+        def call
+          puts "help for generate"
+        end
+      end
+
       class Model
         def call
           puts "generated model"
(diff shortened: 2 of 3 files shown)


spec/support/fixtures/foo
```diff
@@ -30,10 +30,10 @@ module Foo
       end
     end
 
-    register 'hello',     Hello
-    register 'version',   Version
+    register 'hello', Hello
+    register 'version', '-v', '--version', Version
 
-    register 'generate model',  Generate::Model
+    register 'generate model', 'generate --model', Generate::Model
```
jodosha: @AlfonsoUceda Can we keep this realistic? It's probably not the case that someone will define a command like this. What if we leave the line as it is, but instead we add:

```ruby
register 'generate --help', Generate::Help
```

lib/hanami/cli.rb
```diff
@@ -26,8 +26,11 @@ def run(arguments: ARGV)
         true
       end
 
-      def register(name, command)
-        Hanami::Cli.register(name, command)
+      def register(*args)
```
davydovanton: Just a stylistic issue: maybe is better to use two arguments for method definition?
```
def register(*names, command)
  names.each { |name| Hanami::Cli.register(name, command) }
end
```

AlfonsoUceda: @davydovanton better ;)

AlfonsoUceda:
@davydovanton @jodosha I've amended your requets :)

oana-sipos:
Looks good :) Sorry, saw this but somehow did not think I should comment.

PR #9, opened 2017-07-20
Added new features

jodosha:
Main features:

  * Trie based registry
  * Aliases
  * Accept any kind of callable object as command.
  * Only `Hanami::Cli` calls `Kernel.exit`
  * Dependency injection for `ARGV` and `$stdout`. Useful during unit tests: `Hanami::Cli.new(registry).call(args: ["foo"], out: StringIO.new)`
  * Separated strategies for single command banner and subcommands usage

---

Usage: given the `foo` executable

```ruby
#!/usr/bin/env ruby
require "bundler/setup"
require "hanami/cli"

module Foo
  module Commands
    extend Hanami::Cli::Registry

    class Hello < Hanami::Cli::Command
      argument :name, required: true

      def call(name:, **options)
        puts "Hello, #{name}"
      end
    end

    register "hello", Hello, aliases: ["hi"] 
  end
end

Hanami::Cli.new(Foo::Commands).call
```

---

Aliases:

```ruby
#!/usr/bin/env ruby
require "bundler/setup"
require "hanami/cli"

module Foo
  module Commands
    extend Hanami::Cli::Registry

    module Generate
      class App < Hanami::Cli::Command
      end

      class Action < Hanami::Cli::Command
      end
    end
 
    register "generate", aliases: ["g"] do |prefix|
      prefix.register "app",    Generate::App
      prefix.register "action", Generate::Action
    end
  end
end

Hanami::Cli.new(Foo::Commands).call
```

---

Realistic example:

```ruby
#!/usr/bin/env ruby
require "bundler/setup"
require "hanami/cli"

module Foo
  module Cli
    module Commands
      extend Hanami::Cli::Registry

      class Server < Hanami::Cli::Command
        desc "Start Foo server (only for development)"

        option :server,         desc: "Force a server engine (eg, webrick, puma, thin, etc..)"
        option :host,           desc: "The host address to bind to"
        option :port,           desc: "The port to run the server on", aliases: ["-p"]
        option :debug,          desc: "Turn on debug output"
        option :warn,           desc: "Turn on warnings"
        option :daemonize,      desc: "Daemonize the server"
        option :pid,            desc: "Path to write a pid file after daemonize"
        option :code_reloading, desc: "Code reloading", type: :boolean, default: true

        example [
          "                    # Basic usage (it uses the bundled server engine)",
          "--server=webrick    # Force `webrick` server engine",
          "--host=0.0.0.0      # Bind to a host",
          "--port=2306         # Bind to a port",
          "--no-code-reloading # Disable code reloading"
        ]

        def call(options)
          puts "server - #{options.inspect}"
        end
      end
    end
  end
end

Foo::Cli::Commands.register "server", Foo::Cli::Commands::Server
Hanami::Cli.new(Foo::Commands).call
```

```shell
% foo server
server - {:code_reloading=>true}

% foo server --help
Command:
  foo server

Usage:
  foo server

Description:
  Start Foo server (only for development)

Options:
  --server=VALUE                  	# Force a server engine (eg, webrick, puma, thin, etc..)
  --host=VALUE                    	# The host address to bind to
  --port=VALUE, -p VALUE          	# The port to run the server on
  --debug=VALUE                   	# Turn on debug output
  --warn=VALUE                    	# Turn on warnings
  --daemonize=VALUE               	# Daemonize the server
  --pid=VALUE                     	# Path to write a pid file after daemonize
  --[no-]code-reloading           	# Code reloading, default: true
  --help, -h                      	# Print this help

Examples:
  foo server                     # Basic usage (it uses the bundled server engine)
  foo server --server=webrick    # Force `webrick` server engine
  foo server --host=0.0.0.0      # Bind to a host
  foo server --port=2306         # Bind to a port
  foo server --no-code-reloading # Disable code reloading
```

---

/cc @timriley for review.


lib/hanami/cli.rb
```diff
@@ -1,37 +1,66 @@
...
 module Hanami
   class Cli
-    attr_reader :commands
-    attr_reader :renderer
+    require "hanami/cli/version"
+    require "hanami/cli/command"
+    require "hanami/cli/registry"
+    require "hanami/cli/parser"
+    require "hanami/cli/usage"
+    require "hanami/cli/banner"
```
timriley: I'm curious to learn why your policy for `require` lines like these is to nest them within the relevant module. I've always done it at the top of the file, but that's really because "it's always how I've done it". Would love to learn your perspective on this :)

lib/hanami/cli/registry.rb
```diff
@@ -0,0 +1,46 @@
...
+        base.class_eval do
+          @commands = CommandRegistry.new
+        end
+      end
+
+      def register(name, command = nil, aliases: [])
+        if block_given?
+          yield Prefix.new(@commands, name, aliases)
+        else
+          command.command_name = name if Cli.command?(command)
```
timriley: Will this mutate the class-level ivar?

timriley: Perhaps a better way to explain why I think it might be useful to allow instantiation of commands is if you have the same command class but with a little bit of configuration, it can perform 2 different things in your CLI. So in this case you'd want to register it twice with these different options:

```ruby
register "thing",         MyApp::CLI::ThingCommand
register "another_thing", MyApp::CLI::ThingCommand, be_slightly_different: true
```

If we stuck to only registering whole-classes, then the only way to provide this "same behaviour but supporting different config" would be to have the command author split it across two separate classes, which feels like a bit of a hack.

lib/hanami/cli.rb
jodosha: Because if you do at the top of the file, then the first definition of `Hanami::Cli` it isn't by `hanami/cli.rb`, but by something else.

lib/hanami/cli/registry.rb
jodosha: @timriley I'm concerned about this problem too.

We have two ways:

  1. Pass the name to the initializer, but that means we loose the feature to have any callable object as command. Because the following syntax will blow up: `register "foo", ->(*) { ... }`.
  1. Mutate the class ivar only if the command is a `Hanami::Cli::Command`.

---

In my understanding you need for `dry` the ability to register any callable object as a command.

While this is nice-to-have feature, I've got the feeling that people interested into `hanami-cli` will "buy" the entire experience: both the registry **and** the commands.

So if you need it, we'll keep it, otherwise we can assume that all the registered commands are `Hanami::Cli::Command` instances, so then we can pass the name to the initializer.

Please let us to know. 😄

timriley: Hi @jodosha, yeah, I totally agree that people should "buy in" to the whole experience of using hanami-cli, both the registry _and_ the commands. I don't think we need to optimise for someone trying to register objects that don't inherit from `Hanami::Cli::Command`.

(In fact, I wonder if we even want to support registering anonymous procs? It seems like they'd be too devoid of features to fit nicely into an overall CLI - i.e. we can't print any help or usage information for them. What was your thinking for adding them?)

So this makes me think we should really build specifically for the interface of `Hanami::Cli::Command`, and if a user of this gem wants to register commands of a different class, they'd just need to conform to that interface. This means we could indeed initialize command objects as they're registered and leave the flexibility open for the command classes to provide their own initialization logic if they ever needed it.

Here's a diff where I made this happen on top of code in this PR:

```patch
diff --git a/lib/hanami/cli.rb b/lib/hanami/cli.rb
index ef1806e..2a547c6 100644
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -39,10 +39,10 @@ module Hanami
       command = result.command
       return [command, result.arguments] unless command?(command)

-      result = Parser.call(command.class, result.arguments, result.names)
+      result = Parser.call(command, result.arguments, result.names)

       if result.help?
-        Banner.call(command.class, out)
+        Banner.call(command, out)
         exit(0)
       end

diff --git a/lib/hanami/cli/command.rb b/lib/hanami/cli/command.rb
index ce20099..faa71c3 100644
--- a/lib/hanami/cli/command.rb
+++ b/lib/hanami/cli/command.rb
@@ -1,3 +1,4 @@
+require "forwardable"
 require "hanami/cli/option"
 require "concurrent/array"

@@ -11,7 +12,6 @@ module Hanami
       module ClassMethods
         def self.extended(base)
           base.class_eval do
-            @command_name = nil
             @description  = nil
             @examples     = Concurrent::Array.new
             @arguments    = Concurrent::Array.new
@@ -23,7 +23,25 @@ module Hanami
         attr_reader :examples
         attr_reader :arguments
         attr_reader :options
-        attr_accessor :command_name
+      end
+
+      extend Forwardable
+
+      delegate [
+        :description,
+        :examples,
+        :arguments,
+        :options,
+        :params,
+        :default_params,
+        :required_arguments,
+        :optional_arguments,
+      ] => "self.class"
+
+      attr_reader :command_name
+
+      def initialize(command_name:, **)
+        @command_name = command_name
       end

       def self.desc(description)
diff --git a/lib/hanami/cli/command_registry.rb b/lib/hanami/cli/command_registry.rb
index 0d0459e..6e153dd 100644
--- a/lib/hanami/cli/command_registry.rb
+++ b/lib/hanami/cli/command_registry.rb
@@ -7,9 +7,9 @@ module Hanami
         @root = Node.new
       end

-      def set(name, command, aliases)
+      def set(name, command, aliases, **options)
         node = @root
-        command = command_for(command)
+        command = command_for(name, command, **options)
         name.split(/[[:space:]]/).each do |token|
           node = node.put(node, token)
         end
@@ -50,14 +50,14 @@ module Hanami

       private

-      def command_for(command)
+      def command_for(name, command, **options)
         case command
         when NilClass
           command
         when ->(c) { c.respond_to?(:call) }
           command
         else
-          command.new
+          command.new(command_name: name, **options)
         end
       end

diff --git a/lib/hanami/cli/registry.rb b/lib/hanami/cli/registry.rb
index 4897e93..2d9c2f7 100644
--- a/lib/hanami/cli/registry.rb
+++ b/lib/hanami/cli/registry.rb
@@ -9,12 +9,11 @@ module Hanami
         end
       end

-      def register(name, command = nil, aliases: [])
+      def register(name, command = nil, aliases: [], **options)
         if block_given?
           yield Prefix.new(@commands, name, aliases)
         else
-          command.command_name = name if Cli.command?(command)
-          @commands.set(name, command, aliases)
+          @commands.set(name, command, aliases, **options)
         end
       end

@@ -30,11 +29,9 @@ module Hanami
           registry.set(prefix, nil, aliases)
         end

-        def register(name, command, aliases: [])
-          command_name         = "#{prefix} #{name}"
-          command.command_name = command_name if Cli.command?(command)
-
-          registry.set(command_name, command, aliases)
+        def register(name, command, aliases: [], **options)
+          command_name  = "#{prefix} #{name}"
+          registry.set(command_name, command, aliases, **options)
         end

         private
diff --git a/lib/hanami/cli/usage.rb b/lib/hanami/cli/usage.rb
index 1e21f4c..043bf30 100644
--- a/lib/hanami/cli/usage.rb
+++ b/lib/hanami/cli/usage.rb
@@ -35,8 +35,8 @@ module Hanami
       def self.arguments(command)
         return unless Cli.command?(command)

-        required_arguments = command.class.required_arguments
-        optional_arguments = command.class.optional_arguments
+        required_arguments = command.required_arguments
+        optional_arguments = command.optional_arguments

         required = required_arguments.map { |arg| arg.name.upcase }.join(' ') if required_arguments.any?
         optional = optional_arguments.map { |arg| "[#{arg.name.upcase}]" }.join(' ') if optional_arguments.any?
@@ -48,7 +48,7 @@ module Hanami
       def self.description(command)
         return unless Cli.command?(command)

-        " # #{command.class.description}" unless command.class.description.nil?
+        " # #{command.description}" unless command.description.nil?
       end

       def self.justify(string, padding, usage)
```

This gist of this change is:

1. Initialize command object upon registration, passing in the `command_name` at that point, plus any extra kwargs (these may be things that specific command subclasses want to use)
2. Add forwarders for all the methods on the command instances to their respective class methods
3. Stop referring to `command.class` anywhere else in the gem

This means that:

1. The command_name no longer leaks across registrations of the same class
2. Extra options passed at registration time now get passed to the command initializer
3. Command usage is generally more flexible since all the methods the CLI wants to use are on the _instance_, so theoretically, someone could register a completely different class if they wanted, and as long as it responded to the same methods, it's all good
4. We still support registration of anonymous procs (extra options passed to register are just ignored for those)

But TBH I'm totally fine for you to remove anonymous proc registration, I was never planning to use those. I imagine doing this would streamline the code a little: we could simplify `#command_for` and also perhaps stop doing a bunch of calls to `Hanami::Cli.command?` across the place, instead just being confident that the command objects we're working with all implement the proper command interface.

What do you reckon?

AlfonsoUceda:
@jodosha I've seen we don't display alias for subcommands with it. What do you think to print alias in the description?

spec/support/fixtures/foo
```diff
@@ -25,7 +25,7 @@ module Foo
 
       class Console < Hanami::Cli::Command
         desc "Starts Foo console"
-        option :engine, desc: "Force a console engine: (irb/pry/ripl)"
+        option :engine, desc: "Force a console engine", list: %w(irb pry ripl)
```
timriley: @AlfonsoUceda `list:` as the option name here feels a little funny to me... what do you think of something like `values:`?

AlfonsoUceda: @timriley yeah, I think make more sense ;)

jodosha:
@timriley I added you to the committers of this repository. Please feel free to apply the patch above to master.

@AlfonsoUceda Likewise, please make the change of the `list` => `values` in master as well.

---

Just ping me once done, cause I need to reflect the changes in `hanami`. I'm gonna merge this as it is. Thank you all! 💯

PR #11, opened 2017-07-26
Rename option `list:` to `values:`

timriley:
Rename `list:` option for command options to `values:`, since this seems a little more natural, e.g.

```ruby
option :engine, desc: "Force a console engine", values: %w(irb pry ripl)
```


AlfonsoUceda:
cc @jodosha

PR #18, opened 2017-10-14
Added README table of contents (TOC).

bkuhlmann:
## Overview

Improves readability by being able to navigate the documentation with
clickable links. You can move the location of the TOC around in the
document as you like but is best if placed right after the document
label.

At any time, you can regenerate the TOC by running `tocer -g` from the
CLI. Everything else will be handled for you. There is no further
maintenance required. It's a good practice to do this during each
version release.

For more information (or advanced usage) of the Tocer gem, check out
the project [README](https://github.com/bkuhlmann/tocer) for details.

## Screenshots

toc

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,34 @@ General purpose Command Line Interface (CLI) framework for Ruby.
 
 :warning: **This is a general framework for Ruby (aka `thor` gem replacement), NOT the implementation of the `hanami` CLI commands** :warning:
 
+<!-- Tocer[start]: Auto-generated, don't remove. -->
+
+## Table of Contents
+
+  - [Features](#features)
+    - [Registration](#registration)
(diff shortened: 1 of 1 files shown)


bkuhlmann:
:information_source: Looks like the Travis CI build is going to fail as I don't support older Ruby versions within the Tocer gem. An alternative solution would be to remove the `gemspec` development dependency requirement which makes it a bit more inconvenient for future development but not terrible as long as people know to update their dev environment for new versions. I suppose this requirement could be mentioned in the README too.

bkuhlmann:
@jodosha Sounds good. I've fixed (amended) the commit so only the `README` has changed. You should be set.

jodosha:
@bkuhlmann Thanks for understanding. Merging this. Thank you! 👍

PR #27, opened 2018-01-24
Make unused arguments available to the end user

### LINKED ISSUE #26: Variadic Arguments
Hello,
Firstly thanks for Hanami CLI! It is so much more elegant than Thor, which has been my goto tool up to this point.

On to business, I'm in the process of porting an old Thor based CLI to Hanami CLI and I found a dealbreaker missing feature for me - being able to parse and use a variable number of arguments to the CLI.  

The use case here is something similar to `ssh`, `docker run`, and the like where the user provides a series of arguments that'll become the command our program will execute somewhere.

Example:
```bash
ssh root@192.168.0.1 -- /bin/bash -c "echo '127.0.0.1 myapp.com' > /etc/hosts"
```
```bash
docker run -it --rm postgres:latest /bin/bash
```
```bash
docker run -it --rm postgres:latest -- psql postgres://user:pass@myhost/mydb -c "select * from users"
```

As you can see they can get pretty elaborate ;)

I've kind of hacked it in this way - but I'd rather make it official and do it more elegantly:

```patch
diff --git a/lib/hanami/cli/parser.rb b/lib/hanami/cli/parser.rb
index e9b7534..2cc18b1 100644
--- a/lib/hanami/cli/parser.rb
+++ b/lib/hanami/cli/parser.rb
@@ -27,6 +27,8 @@ module Hanami
           end
         end.parse!(arguments)

+        Hanami::CLI.define_singleton_method(:all_arguments) { arguments }
+
         parsed_options = command.default_params.merge(parsed_options)
         parse_required_params(command, arguments, names, parsed_options)
       rescue ::OptionParser::ParseError
```

I'm happy to do the work on it but figured I'd ask first as to how it should be done - any opinions?

whoward:
In several use cases users will want to allow their CLI users to pass in
a variable number of arguments to the CLI.

Example:

```bash
ssh root@192.168.0.1 -- /bin/bash -c "echo '127.0.0.1 myapp.com' >
/etc/hosts"

docker run -it --rm postgres:latest /bin/bash

cat foo.txt bar.txt baz.txt
```

This commit allows users to process unused arguments by assigning them
to `Hanami::CLI.unused_arguments`

Closes #26

### DIFF
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -13,6 +13,20 @@ module Hanami
     require "hanami/cli/usage"
     require "hanami/cli/banner"
 
+    class << self
+      # Returns a list of arguments have not been defined for the subcommand.
+      #
+      # @return [Array] list of unused arguments
+      # @since 0.2.0
+      attr_reader :unused_arguments
+
+      # Assigns the unused arguments from the parser.
+      #
+      # @since 0.2.0
+      # @api private
+      attr_writer :unused_arguments
+    end
+
     # Check if command
     #
     # @param command [Object] the command to check
--- a/lib/hanami/cli/parser.rb
+++ b/lib/hanami/cli/parser.rb
@@ -51,6 +51,8 @@ module Hanami
         parse_required_params = Hash[command.required_arguments.map(&:name).zip(arguments)]
         all_required_params_satisfied = command.required_arguments.all? { |param| !parse_required_params[param.name].nil? }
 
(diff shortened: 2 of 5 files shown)


AlfonsoUceda:
@whoward thank you so much for this PR, I'll test as soon as possible but it seems good ;)

AlfonsoUceda:
@whoward it works well, tested manually and with a hanami branch, can you add a test using an option too? after that I can merge it ;)

AlfonsoUceda:
@whoward, don't worry, I'm going to add it myself

PR #28, opened 2018-02-06
Introduce callbacks for CLI commands

davydovanton:
From the `foo` gem we have a command `hello`.

```ruby
#!/usr/bin/env ruby
require "hanami/cli"

module Foo
  module CLI
    module Commands
      extend Hanami::CLI::Registry

      class Hello < Hanami::CLI::Command
        argument :name, required: true

        def call(name:, **)
          puts "hello #{name}"
        end
      end
    end
  end
end

Foo::CLI::Commands.register "hello", Foo::CLI::Commands::Hello

cli = Hanami::CLI.new(Foo::CLI::Commands)
cli.call
```

The `foo-bar` gem enhances `hello` command with callbacks:

```
Foo::CLI::Commands.before("hello") { |args| puts "debug: #{args.inspect}" } # syntax 1
Foo::CLI::Commands.after "hello", &->(args) { puts "bye, #{args.fetch(:name)}" } # syntax 2
```

```shell
% foo hello Anton
debug: {:name=>"Anton"}
hello Anton
bye, Anton
```


spec/support/fixtures/foo
```diff
@@ -463,6 +463,9 @@ module Foo
       end
 
       class SubCommand < Hanami::CLI::Command
+        before { puts 'Before sub command' }
```
jodosha: I'd not add this as DSL for a command.
The point of callbacks feature is to extend somebody else's commands. We have two cases:

  * If you are the author of this command, you're writing this concrete class. There is no need to have a callback as you can implement all the logic in `#call`.
  * You are a person who want to extend this command, you can't access `Subcommand.before`, because that requires to monkey-patch the original class.

What we can do is:

```ruby
# gem `foo`

module Foo
  module CLI
    class Hello < Hanami::Cli::Command
      def call(*)
        puts "Hello"
      end
    end
  end
end

Foo::CLI.register "greet", Foo::CLI::Hello
```

Then in the second gem (`bar`)

```ruby
# gem `bar`
require "foo"

module Bar
  module CLI
    class World < Hanami::Cli::Command
      def call(*)
        puts "world!"
      end
    end
  end
end

Foo::CLI.after "greet", Bar::CLI::World
```

And finally in the shell:

```shell
% greet
Hello
world!
```

davydovanton: yep, I'll do it. My next step (after specs and cleanup) - update command resolver to call `before` and `after` methods for each command with new "callback" command

spec/support/fixtures/foo
```diff
@@ -477,5 +480,11 @@ Foo::CLI::Commands.register "generate webpack", Foo::Webpack::CLI::Generate
 Foo::CLI::Commands.register "hello",            Foo::Webpack::CLI::Hello
 Foo::CLI::Commands.register "sub command",      Foo::Webpack::CLI::SubCommand
 
+# we need to be sure that command will not override with nil command
+Foo::CLI::Commands.register "generate webpack", nil
+
+Foo::CLI::Commands.after("sub command") { puts 'After command hook' }
+Foo::CLI::Commands.before("sub command") { puts 'before command hook' }
```
davydovanton: @jodosha finally I got something like this. WDYT about API? Now I want to cleanup all code and add specs

lib/hanami/cli/command.rb
```diff
@@ -13,6 +14,7 @@ class Command
       def self.inherited(base)
         super
         base.extend ClassMethods
+        base.include Callbacks
```
jodosha: @davydovanton There is no need to allow callbacks DSL to a command.

Again, we **won't** allow this syntax:

```ruby
class Command < Hanami::CLI::Command
  before { ... }
  after { ... }
end
```

Including this `Callbacks` module here, will create public API that we **won't** support and need.

lib/hanami/cli/registry.rb
```diff
@@ -86,6 +86,24 @@ def get(arguments)
         @commands.get(arguments)
       end
 
+      # @since x.x.x
+      # @api private
+      def before(command, &callback)
+        command_class(command).before(&callback)
```
jodosha: We can't delegate to a class the callbacks. If a command gets replaced, the attached callbacks won't be executed.

Let's say we have a `foo` gem with a CLI:

```ruby
module Foo
  module CLI
    module Commands
      extend Hanami::CLI::Registry

      class Console < Hanami::CLI::Command
        # ...
      end
    end
  end
end

Foo::CLI::Commands.register "console", Foo::CLI::Commands::Console
```

Then we have a `foo-greeting` gem, that provides a callback:

```ruby
Foo::CLI::Commands.after "console", { puts "goodbye" }
```

With this last operation, we have attached the callback to `Foo::CLI::Commands::Console`.

Now a third gem is loaded: `foo-pry`

```ruby
module Foo
  module Pry
    class Console < Hanami::CLI::Command
      # ...
    end
  end
end

Foo::CLI::Commands.register "console", Foo::Pry::Console
```

We have replaced the object that is invoked when we invoke `foo console` from shell.

But because callbacks were attached to the old class, they won't be executed.

lib/hanami/cli/registry.rb
```diff
@@ -86,6 +86,24 @@ def get(arguments)
...
+      end
+
+      # @since x.x.x
+      # @api private
+      def after(command, &callback)
+        command_class(command).after(&callback)
```
jodosha: What I suggest to do is to register callbacks to `CommandRegistry`, which is referenced as `@commands` from this file.

Then both `CommandRegistry#get` and `#set` should manage callbacks.

davydovanton: nice catch 💯

davydovanton:
@jodosha thank you for really great feedback 💯 

-------------------

### My next iteration

I replace callback code to CommandRegistry class (Node).

Also, I update `foo` fixture (from specs) code to:
```ruby
Foo::CLI::Commands.register "sub command",      Foo::Webpack::CLI::SubCommand

Foo::CLI::Commands.after("sub command") { puts 'After command hook' }
Foo::CLI::Commands.before("sub command") { puts 'before command hook' }
```

And I have this result:
screenshot 2018-02-22 23 50 08


After updating `foo` file to:
```ruby
Foo::CLI::Commands.register "sub command",      Foo::Webpack::CLI::SubCommand

Foo::CLI::Commands.after("sub command") { puts 'After command hook' }
Foo::CLI::Commands.before("sub command") { puts 'before command hook' }

Foo::CLI::Commands.register "sub command",      Foo::Webpack::CLI::Hello
```

I get this result:
screenshot 2018-02-22 23 50 19

--------------------


WDYT about current realisation logic?

davydovanton:
@jodosha let's merge it and start working on next part 💯 

@AlfonsoUceda could you check it?

davydovanton:
I merged it, will start work on next task 👍

PR #33, opened 2018-04-13
Add required_ruby_version to gemspec

dogenpunk:
This change leads up to a fix for hanami/hanami#924

### DIFF
--- a/hanami-cli.gemspec
+++ b/hanami-cli.gemspec
@@ -27,6 +27,8 @@ Gem::Specification.new do |spec|
     f.match(%r{^(test|spec|features)/})
   end
 
+  spec.required_ruby_version = ">= 2.4.0"
+
   spec.add_dependency "hanami-utils",    "~> 2.0.alpha"
   spec.add_dependency "concurrent-ruby", "~> 1.0"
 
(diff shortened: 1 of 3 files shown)


hanami-cli.gemspec
```diff
@@ -27,6 +27,8 @@ Gem::Specification.new do |spec|
     f.match(%r{^(test|spec|features)/})
   end
 
+  spec.required_ruby_version = ">= 2.4.0"
```
davydovanton: hanami cli works with jruby too. How we can handle it here?

dogenpunk: That's a good question. I'm trying to get up to speed on jruby support. The other hanami gems seem to use this setting already so I thought it was an oversight here. I'll come back to this after I've completed this process for them.

AlfonsoUceda: Folks, this is related to jruby 9.1 because it targets ruby 2.3, jruby 9.2 will target ruby 2.5

davydovanton:
agree with @AlfonsoUceda 👏

AlfonsoUceda:
can you update to jruby 9.2.0.0? it was release some days ago. thanks!

AlfonsoUceda:
@dogenpunk can you update travis ci file to reference jruby 9.2.0.0? thanks!

IvanShamatov:
Closing this one as outdated. The initial problem was solved in hanami

PR #38, opened 2018-05-11
Extract generator from hanami gem

cllns:
This is extracted from `hanami/hanami`. See https://github.com/hanami/hanami/pull/931, which relies on this change.

Tests still need to be written before this can be merged, but I wanted to get this up so people can see the changes I made.

The idea is to abstract away the file operations and also `say` to the `out` stream what's happening. This way, we can alter the output (to colorize it) and also catch and handle any errors (like overwriting an existing file).


davydovanton:
awesome, really want to see this changes 👏 

_Also, small off topic, @cllns @AlfonsoUceda @jodosha WDYT about user testing helpers for generators?_

lib/hanami/cli/generator.rb
```diff
@@ -0,0 +1,128 @@
...
+module Hanami
+  class CLI
+    # A class to help generate files
+    #
+    # @since 0.1.0
+    class Generator
+      # @since x.x.x
+      # @api public
+      def initialize(out: $stdout, files: Utils::Files)
+        @files     = files
```
AlfonsoUceda: extra spaces?

lib/hanami/cli/generator.rb
```diff
@@ -0,0 +1,128 @@
...
+
+      # @since x.x.x
+      # @api public
+      def delete(path, allow_missing: false)
+        return if allow_missing && !files.exist?(path)
+        files.delete(path)
+        say(:remove, path)
+      end
+
+      # TODO: add allow_missing?
```
AlfonsoUceda: Yes, we can allow that to avoid raise an error, isn't it?

lib/hanami/cli/generator.rb
```diff
@@ -0,0 +1,128 @@
...
+      # @api private
+      class Renderer
+        TRIM_MODE = "-".freeze
+
+        def initialize
+          freeze
+        end
+
+        def call(template, context)
+          ::ERB.new(template, nil, TRIM_MODE).result(context)
```
AlfonsoUceda: should we support more templates? I don't think so, it's ok to me.

lib/hanami/cli/generator.rb
```diff
@@ -0,0 +1,128 @@
...
+      end
+
+      # @since x.x.x
+      # @api public
+      def create(source, destination, context)
+        files.write(destination, render(source, context))
+        say(:create, destination)
+      end
+
+      # Q: Should we replace this method with `create` and passing in
```
AlfonsoUceda: No, we shouldn't, let's keep this I is ok say just touch :)

lib/hanami/cli/generator.rb
```diff
@@ -0,0 +1,146 @@
...
+
+      private
+
+      # @since x.x.x
+      # @api private
+      SAY_FORMATTER = "%<operation>12s  %<path>s\n".freeze
+
+      # @since x.x.x
+      # @api private
+      class Renderer
```
cllns: I'm not sure we need this class? Seems like it could just be a helper method.

It was taken from: https://github.com/hanami/hanami/blob/master/lib/hanami/cli/commands/command.rb#L105-L125

lib/hanami/cli/generator.rb
```diff
@@ -0,0 +1,146 @@
...
+
+module Hanami
+  class CLI
+    # A class to help generate files
+    #
+    # @since 0.1.0
+    class Generator
+      # @since x.x.x
+      # @api public
+      def initialize(out: $stdout, files: Utils::Files, templates_dir: __dir__)
```
cllns: I'm no sure if `__dir__` is a good default here. The other option is that we let it be `nil`, then raise a helpful error if someone tries to use `create` or `copy` without passing this argument in.

I'm also not sure `templates_dir` is the right name for this variable. For `create`, it's the right name, but for `copy`, the files are necessarily templates. For example, in `hanami/hanami`, we copy over a `favicon.ico`, which isn't a template at all. Maybe `root_dir` would be better?

cllns:
I extended this PR to also be able to find its own templates, by passing in an argument of where they're located. For context, it's to allow the behavior described in this comment on `hanami/hanami`: https://github.com/hanami/hanami/pull/931#issuecomment-403648998

Still need to write specs for this class. The specs from `hanami/hanami` in `spec/integration/cli/generate/` pass locally using this code, so it's tested, just not in isolation. (Waiting on full suite pass)

cllns:
@AlfonsoUceda @davydovanton @jodosha Can you give me some feedback on this? 

`Generator` didn't seem like the right name anymore since we extracted it, so I renamed it to `FileHelper`. But, I'm not sure `FileHelper` is the best name since it's not _only_ for files, it can also execute arbitrary shell commands with `#exectue`. 

For `hanami/hanami`, we only use it to run `git init`, so we could remove the all-purpose `#execute` in favor a more specialized `#initialize_git` method, which _is_ related to files. Still though, in terms of this being a multi-purpose utility, I image other users of this library might want to run arbitrary commands

I added specs but mostly just for the happy path, not many sad paths. What should we do if a file isn't found? I didn't add tests here, but `Errno::ENOENT` is the main Exception we can expect from `Hanami::Utils::Files`, and it would propagate up. We could rescue that and give a more helpful error, but I'm not sure if that's unnecessarily obscuring what's happening.

Any ideas or thoughts here? :)

You can see the associated changes for `hanami/hanami` in this PR: https://github.com/hanami/hanami/pull/931

lib/hanami/cli/generator.rb
jodosha: @cllns The problem with `root_dir` it can be confused with the root of a project (eg. `Hanami.root`).

cllns: :+1: It's already been changed to `templates_dir`

timriley:
@cllns Regarding the best naming for this collection of methods, what do you think about splitting them up into `FileHelper` and `ShellHelper` modules? This would allow you to move `#execute` into `ShellHelper` and avoid the issue of the `FileHelper` name not being fully representative of the methods it contains. It'd also offer a natural place for other shell helper methods to be added (I can think of at least one, where there's a long-running shell command to be executed and we want to show a spinner or something until it completes).

jodosha:
@cllns Can you please share with us an example of usage of this module? Because that seems to be pretty solid self-contained piece of code, but I'm struggling to see it in the complete picture.

cllns:
After chatting with @IvanShamatov, he will be integrating these features into `dry-cli` in a different way, so I'm closing this PR.

cllns:
This exists now as https://github.com/dry-rb/dry-files

PR #40, opened 2018-06-21
Implement variadic arguments in a thread-safe way

AlfonsoUceda:
## Feature

Sometimes we need extra arguments because those will be forwarded to a sub-command like `ssh`, `docker` or `cat`.

By using `--` (double dash, aka hypen), the user indicates the end of the arguments and options belonging to the main command, and the beginning of the variadic arguments that can be forwarded to the sub-command.
These extra arguments are included as `:args` in the keyword arguments available for each command.

```ruby
#!/usr/bin/env ruby
require "bundler/setup"
require "hanami/cli"

module Foo
  module CLI
    module Commands
      extend Hanami::CLI::Registry

      class Runner < Hanami::CLI::Command
        argument :image, required: true, desc: "Docker image"

        def call(image:, args: [], **)
          puts `docker run -it --rm #{image} #{args.join(" ")}`
        end
      end

      register "run", Runner
    end
  end
end

Hanami::CLI.new(Foo::CLI::Commands).call
```

```shell
% foo run ruby:latest -- ruby -v
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
```

The user separates via `--` the arguments for `foo` and the command has to be run by the Docker container.
In this specific case, `ruby:latest` corresponds to the `image` mandatory argument for `foo`, whereas `ruby -v` is the variadic argument that is passed to Docker via `args`.

---

This is a revamp of PR #29.

We reverted PR #27 because it wasn't thread-safe so with this new PR we bring thread-safety to variadic arguments.

### DIFF
--- a/lib/hanami/cli/parser.rb
+++ b/lib/hanami/cli/parser.rb
@@ -51,6 +51,8 @@ module Hanami
         parse_required_params = Hash[command.required_arguments.map(&:name).zip(arguments)]
         all_required_params_satisfied = command.required_arguments.all? { |param| !parse_required_params[param.name].nil? }
 
+        unused_arguments = arguments.drop(command.required_arguments.length)
+
         unless all_required_params_satisfied
           parse_required_params_values = parse_required_params.values.compact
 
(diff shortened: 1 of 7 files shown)


lib/hanami/cli/parser.rb
```diff
@@ -64,7 +66,7 @@ def self.parse_required_params(command, arguments, names, parsed_options)
         end
 
         parse_params.reject! { |_key, value| value.nil? }
-        Result.success(parsed_options.merge(parse_params))
+        Result.success(parsed_options.merge(parse_params).merge(unused_arguments: unused_arguments))
```
AlfonsoUceda: what do you think to add only unused_arguments if the array has elements?

Issue #47, opened 2018-12-06
Return non-zero exit status


fernandes:
Hi,

If I create a command and would like to automate the use of CLI, is there any way to return a non-zero exit status?

I tried the most intuitive way:

```ruby
def call(*)
  do_something
  return 3
end
```

but doesn't work, any tip on this? 😉

davydovanton:
hey, could you explain the problem? Why do you want to return specific data from CLI command and what (and where) do you expect to do with this data?

fernandes:
hey @davydovanton, first thank you for your reply, sure here we go..

unix consider exit code as 0 == run was ok, > 0 if any problem happened (maybe I'm not explaining correctly, here are the more details on [wikipedia](https://en.wikipedia.org/wiki/Exit_status#POSIX))

resuming, the idea is:

```bash
my_cli comand_will_succeed
echo $?
=> 0

my_cli command_will_fail
echo $?
=> X (X > 0, defined by the programmer)
```

so when automating the CLI call (using a shell script, puppet, whatever) can check if it was a success or not only by the exit code

thank you!

davydovanton:
got the point, thanks!

EgorBazhenov:
It seems to me that to solve this problem, we can try to return an object-result from the commands, which would know the return value and the exit-status.
I am not sure that this decision is worth it: major, rather bold revision.

If we trying to look at this issue in the general case, the problem is in exception-flow control, which may be like this:

```Ruby
def call
  # command code here
rescue SomeException => e
  logger.log_exception(e)
  exit(e.exit_code)
end
```
But in current data-flow, there are a callbacks, which can also raise an exceptions. So, here is my suggestion: add custom block to control exception-flow on whole chain `before -> main -> after`. I can make it, if this solutions looks normal.

IvanShamatov:
Closing this one since PR is present, the continued conversation is in #48

PR #48, opened 2018-12-06
Returns an exit code (default 0 for success or exit code) based on command return

fernandes:
This PR is just a kickstart to discuss https://github.com/hanami/cli/issues/47 (motivation is on https://github.com/hanami/cli/issues/47#issuecomment-444873040)

I'm pretty sure just the return of command is not a good solution, AND it should be more reliable/explicit... maybe set the return with a specific method on the `Command` and not rely only on the return of the `call` method

Anyway, I'm just opening this PR so we can discuss if it worths to implement, and the best design for the problem.

ps: I isolated the test on a separated file so would be easy to check and discuss it, I tried to touch the current implementation as few as possible

thank you!

### DIFF
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -54,9 +54,8 @@ module Hanami
         command, args = parse(result, out)
 
         result.before_callbacks.run(command, args)
-        command_ret = command.call(args)
+        command_ret = command.call(args) || 0
         result.after_callbacks.run(command, args)
-        command_ret = 0 if command_ret.nil?
         exit(command_ret)
       else
         usage(result, out)
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -54,8 +54,10 @@ module Hanami
         command, args = parse(result, out)
 
         result.before_callbacks.run(command, args)
-        command.call(args)
+        command_ret = command.call(args)
         result.after_callbacks.run(command, args)
+        command_ret = 0 if command_ret.nil?
+        exit(command_ret)
       else
         usage(result, out)
       end
--- a/spec/integration/exit_code_spec.rb
+++ b/spec/integration/exit_code_spec.rb
@@ -1,0 +1,13 @@
+RSpec.describe "Commands" do
+  it "exit with standard 0 error code" do
+    `foo exit_code valid`
+
+    expect($?.exitstatus).to eq(0)
+  end
+
+  it "exit with 1 error code when fail" do
+    `foo exit_code invalid`
+
+    expect($?.exitstatus).to eq(1)
+  end
+end
--- a/spec/integration/exit_code_spec.rb
+++ b/spec/integration/exit_code_spec.rb
@@ -1,13 +1,15 @@
+require 'English'
+
 RSpec.describe "Commands" do
   it "exit with standard 0 error code" do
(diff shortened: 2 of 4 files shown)


lib/hanami/cli.rb
```diff
@@ -54,8 +54,9 @@ def call(arguments: ARGV, out: $stdout)
         command, args = parse(result, out)
 
         result.before_callbacks.run(command, args)
-        command.call(args)
+        command_ret = command.call(args) || 0
```
paulodiovani: I think this is not a good solution.

Return values are intended to pass messages among methods. Exit codes are a completely different thing.

In bash, for example, return values can be piped, while exit codes are usually used in conditionals:

```bash
# return value example
echo 'foo-bar' | tr '-' ' '

# exit code example
command -v curl && curl http://example.com/foobar || echo command `curl` does not exist
```

My suggestion is:

1. Whenever a method complete successfully, it just resolves normally
    + the resulting command returns 0
    + the method can return whatever value it is supposed to, which may be used or not

2. When a method fails anyway, it should throw an error
    + instead of relying on the return value, the cli should rescue the error, if any
    + the nature of the error (the erro class) is used to decide which exit code to use. General errors shall use code 1. 
          See https://www.tldp.org/LDP/abs/html/exitcodes.html for examples.

fernandes: hey @paulodiovani thank you for reply

I don't think this is a good solution too, it was just a PoC to show the problem I was trying to solve considering #47

My idea was exactly to start a conversation, so I do appreciate your solution proposal, the only point is that exit status should **not** be defined by hanami-cli, instead, must be defined by the user... because it's command related, not framework related

thank you again

paulodiovani: > exit status should **not** be defined by hanami-cli

Not sure. The cli will be the interface with the tty (terminal), that's why it's what should decide which error code to exit. The implemented method, OTOH, just have to throw an exception, so a different interface (e.g. a web service) could use the same implementation while exposing the errors in another way.

fernandes: > The cli will be the interface with the tty (terminal), that's why it's what should decide which error code to exit.

we disagree in this point heheh

was just a suggestion, no worries, as I'm automating using a shell script, I need to control the flow using exit code, but again, no worries, I can find a solution myself 😉

paulodiovani: Just found that any non-rescued exception is already returning a status `!= 0`.

So, in short, you just want to decide which error code is used, right? :thinking:

IvanShamatov:
@fernandes Hello there, I would love to return to this PR, if you don't mind.
I agree with 

> The cli will be the interface with the tty (terminal), that's why it's what should decide which error code to exit.

Could you please clarify, why don't you just call `exit` in your command directly, if you want custom exit code?
```ruby
class Alpha < Dry::CLI::Command
  def call(*); exit(188); end
end
```
That's what I've after running this one:
```shell
$ echo $?
188
```

IvanShamatov:
Closing this PR. CLI doesn't limit you to use your exit codes. If you need any, just finish your command by exiting with special status.

fernandes:
hello @IvanShamatov , yeah, I don't know why I haven't tried it before hahah but it worked, so makes sense to close the PR and just call `exit`, thank you!

PR #54, opened 2019-07-14
Possibly better handling of exit code & unknown command reporting

kigster:
### Context

This PR fixes a couple of issues with Hanami::CLI:

 * When command is run without arguments, it prints out help. And yet, it used to exit with a non-zero status. This PR fixes that — now when help is printed, the command exits with zero status.

 * When `hanami-cli` is used by another library which also uses Aruba's `InProcess` runner, hardcoded calls to `exit()` terminate tests before they can finish. Now one can pass a double for the `Kernel` module, and avoid that.

 * When a subcommand is passed which was not recognized, no error was generated. Given a CLI tool that uses this library and has many commands and subcommands, a user may be confused as to what command line argument failed to map to a known subcommand and resulted in an error (expressed as a help message shown). Now, if the user types a command thats not valid, it is reported to `$stderr`. This behavior can be overridden by passing `nil` for `err:`, for example:

```ruby
cli = Hanami::CLI.new(Foo::CLI::Commands)
cli.call(err: nil) # will not print unknown command errors
cli.call(err: $stdout) # will print unknown command errors to STDOUT
cli.call # will print unknown command errors to STDERR
```

 * I've also added a specific command-exit testing spec that tests both error output and exit status of various combinations.

Thank you so much! I really like this gem, and hope this PR passes the bar and gets merged at some point. I believe should not break any existing usages of `Hanami::CLI`.

### Examples

Given this slightly modified example from the README:

```ruby
#!/usr/bin/env ruby
# File: configure.rb
require "bundler/setup"
require "hanami/cli"

module Foo
  module CLI
    module Commands
      extend Hanami::CLI::Registry

      module Generate
        class Configuration < Hanami::CLI::Command
          def call(*)
          end
        end

        class Project < Hanami::CLI::Command
          def call(*)
          end
        end
      end
    end
  end
end

Foo::CLI::Commands.register "generate configuration", Foo::CLI::Commands::Generate::Configuration
Foo::CLI::Commands.register "generate project", Foo::CLI::Commands::Generate::Project

Hanami::CLI.new(Foo::CLI::Commands).call
```

Assuming this file is called `configure.rb`, previously the following would be the behavior:

**Old Behavior**:

```bash
$ ./configure.rb
Commands:
  configure.rb generate [SUBCOMMAND]
$ echo $?
1
$ ./configure.rb generate asdlkfjasdk
Commands:
  configure.rb generate configuration
  configure.rb generate project
```

**New Behavior**

```bash
$ ./configure.rb
Commands:
  configure.rb generate [SUBCOMMAND]
$ echo $?
0
$ ./configure.rb generate asdlkfjasdk
Error:
  Unknown command(s): asdlkfjasdk

Commands:
  configure.rb generate configuration
  configure.rb generate project
```


.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea/
```
cllns: Can you remove this?

.rubocop.yml
```diff
@@ -2,3 +2,11 @@
 # alphabetically
 inherit_from:
   - https://raw.githubusercontent.com/hanami/devtools/master/.rubocop.yml
+
+AllCops:
+  TargetRubyVersion: 2.3
+  Exclude:
+    - 'spec/support/fixtures/foo'
```
cllns: Could we make the fixture adhere to the style guide instead of doing this?

.rubocop.yml
```diff
@@ -2,3 +2,11 @@
...
+
+Style/SpecialGlobalVars:
```
cllns: Hm so it looks like the alternative requires adding [`require "English"](https://github.com/rubocop-hq/rubocop/issues/1747) to the top of files that use the special global vars. That's pretty weird so I'm not sure which we'd prefer. We might want to move this to `hanami/devtools`'s `.rubocop.yml` but for now I think it's OK, since it might be the only place this is an issue.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: ruby
 sudo: false
-cache: bundler
+#cache: bundler
```
cllns:
```suggestion
cache: bundler
```

hanami-cli.gemspec
```diff
@@ -30,4 +30,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "bundler", ">= 1.6", "< 3"
   spec.add_development_dependency "rake",  "~> 12.0"
   spec.add_development_dependency "rspec", "~> 3.7"
+  spec.add_development_dependency "rubocop"
+  spec.add_development_dependency "rainbow", '=2.2.2'
```
cllns:
```suggestion
```
Unless there's some reason we need this gem / version.

lib/hanami/cli.rb
```diff
@@ -45,27 +45,57 @@ def initialize(registry)
...
-        usage(result, out)
+        exit_code = 0
+        invalid_commands = original_argv.reject { |a| a.start_with?('-') }
+        if invalid_commands.size.positive?
+          err&.puts("Error:\n")
+          err&.puts("  Unknown command(s): #{invalid_commands.join(', ')}")
+          err&.puts
+          exit_code = 1
+        end
+        usage(result, out, exit_code)
```
cllns: Thoughts on extracting this to a helper method? Might help us avoid disabling `Metrics/AbcSize` and `Metrics/MethodLength` too.

cllns:
Great PR, thanks 🌸  I like the dependency injection changes :100:

I've requested some small changes. I'll also wait for another core team member's review on this, but it seems good to me :)

.gitignore
kigster: Why? What's the harm?

timriley: This is the kind of thing you could exclude in your own personal `~/.gitignore`, FWIW :)

.rubocop.yml
kigster: I am doing it now..

kigster:
Thanks for all the feedback! I've submitted another commit that addresses most of the asks. In particular, I was able to rollback ALL of the rubocop changes.

But please let me explain where they came from in the first place.

### Rubocop and TargetRuby Version Madness

Take a look at [this Travis build](https://travis-ci.org/hanami/cli/jobs/558444346) — you'll notice it failed with the following error (just like another 20 builds where I tried various things).

TargetRubyVersion

It took me hours to track this down, and honestly I wasn't a big fan of changing `.rubocop.yml` myself — while I do have some style opinions, I usually keep them to myself until after I submitted about 10 PRs  per project :) 

But this error was perplexing and it was preventing from submitting this PR, so I kept digging.

### Eventual Semi-Discovery

I call this a semi-discovery because I only partially figured this out.

It looks like the devtools's `.rubocop.yml` file always had `TargetRubyVersion 2.3`, in fact it was [reverted in this commit from 2.4](https://github.com/hanami/devtools/commit/fa321addf3ed6bdaa2196fe9c5225fb5261cf202).

But, what I believe has happened is that — rubocop's depends on another gem `rainbow` (ascii coloring gem), and that gem got upgrade to version 3.0.0, and very recently also *added* their `TargetRubyVersion` to 2.3 in [this commit](https://github.com/sickill/rainbow/commit/89b8dcd90121a2578a8d8c002edd7fbf991b433b).

Now, what I do NOT understand is which gem that rubocop depends still requires 2.1 support — I can't see any other explanation, but perhaps one of the authors here might figure it out.

So the *simplest* single line solution to this madness was to fix the `rainbow` gem to a version prior to this "fix", because something in the dependencies somehow requires supporting 2.1.

Rubocop is very happy to depend on a wide range of `rainbow`, so no other changes were necessary to make CI pass. But once Rubocop bumps up `rainbow` again, this may rear it's ugly head.

#### Open Questions

 * What still depends on Ruby 2.1? 
 * Is it a gem, a gem's rubocop's TargetVersion, Travis's matrix, etc?

At the very least with these changes the build is green :)

Thanks again!

.gitignore
kigster: I actually wasn't aware of the personal `~/.gitignore`, thanks!

hanami-cli.gemspec
kigster: See below.

.gitignore
kigster: Done, thanks for the tip!

lib/hanami/cli.rb
```diff
@@ -45,27 +45,65 @@ def initialize(registry)
...
+      err&.puts("  Unknown command(s): #{invalid_commands.join(', ')}")
+      err&.puts
+    end
+
+    # Calls the command and runs callbacks.
+    #
+    # @param result [Hanami::CLI::CommandRegistry::LookupResult]
+    # @param out [IO] sta output
+    #
+    # @since 0.1.0
```
cllns:
```suggestion
    # @since x.x.x
```
The version will be filled in automatically if we do this

lib/hanami/cli.rb
```diff
@@ -45,27 +45,65 @@ def initialize(registry)
...
+    # @api private
+    def call_command(result, out)
+      command, args = parse(result, out)
+
+      result.before_callbacks.run(command, args)
+      command.call(args)
+      result.after_callbacks.run(command, args)
+    end
+
+    # @since 0.1.0
```
cllns:
```suggestion
    # @since x.x.x
```

cllns:
I don't really understand why our CI script is looking at the `.rubocop.yml` files of dependencies. I asked the `rainbow` team to release a `3.0.1` or `3.1.0` with 2.3 as the TargetRubyVersion (see reference above in GitHub UI). This should let us remove that specific version `development_dependency` from the gemspec.

spec/integration/helpers/constants.rb
```diff
@@ -0,0 +1,20 @@
+require 'english'
```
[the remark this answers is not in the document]
kigster: We need it in one place, where we use `$CHILD_STATUS`

kigster:
@cllns Just curious what's the status of this ticket and whether I should do something else to make this PR mergeable?

kigster:
I just realized and fixed a small bug that the PR introduced:

Given this file:

```ruby
#!/usr/bin/env ruby
require "bundler/setup"
require "hanami/cli"

module Foo
  module CLI
    module Commands
      extend Hanami::CLI::Registry

      module Generate
        class Configuration < Hanami::CLI::Command
          def call(*)
          end
        end

        class Project < Hanami::CLI::Command
          def call(*)
          end
        end
      end
    end
  end
end

Foo::CLI::Commands.register "generate configuration", Foo::CLI::Commands::Generate::Configuration
Foo::CLI::Commands.register "generate project", Foo::CLI::Commands::Generate::Project

Hanami::CLI.new(Foo::CLI::Commands).call
```

When we run the `generate` command in help mode, instead of printing this:

```bash
$ ./t.rb generate --help
Commands:
  t.rb generate configuration
  t.rb generate project
```

it printed this:

```bash
$ ./t.rb generate --help
Error:
  Unknown command(s): generate

Commands:
  t.rb generate configuration
  t.rb generate project
```

This has **has now been fixed.**

Thanks!

kigster:
@cllns OK I've updated the PR's description, added an additional test, and fixed a small bug with subcommands. Would really love to hear your thoughts on whether this has a chance to make it into the main codebase or not. Thank you!

IvanShamatov:
Hey, @kigster. I just want to mention, that I've added stderr to the CLI. 
But using CLI with `-h` still returns status 0, as it is valid behavior.
Thanks for your contribution, as it is your PR, which pushed me to dive into logic and learn Unix standards for IO streams. I'm closing this PR.

PR #59, opened 2019-12-07
Adding Hanami::Utils::Files inside Dry::CLI

IvanShamatov:
This is the second step to Dry-CLI with a usage of hanami/utils/files.
I'll make a separate PR to make rubocop happy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ Or install it yourself as:
     - [Aliases](#aliases)
     - [Subcommand aliases](#subcommand-aliases)
     - [Callbacks](#callbacks)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -616,6 +617,14 @@ hello Anton
 bye, Anton
 ```
 
+### File Utilities
+
+File utilities is a set of useful methods to manupulate files and directories, which is opted to be required. [API doc](http://www.rubydoc.info/gems/dry-cli/Dry/CLI/Utils/Files)
```
solnic: `manupulate` => `manipulate`

PR #64, opened 2019-12-14
Removing hanami-utils from dependencies

IvanShamatov:
So there were only 4 usages of hanami/utils/something in the project:
1. *hanami/utils/deprecation* — that is what I forgot to remove when removed the deprecated error
2. *hanami/utils* to use `require!` — replaced with 3-liner to require all files in folders
3. *hanami/utils/string* — both of usages of `dasherize` —  replaced with `Dry::Inflector.new.daserize` calls. Default config of inflector works perfectly against current specs and actually instantiate only 2 objects at the runtime.
4. *hanami/utils/callbacks* — replaced with fully compatible, but small implementation of Chain. It implements only 2 methods of simplified `append` and `run` from `Callbacks::Chain`.

### DIFF
--- a/dry-cli.gemspec
+++ b/dry-cli.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |spec|
   end
 
   spec.add_dependency 'concurrent-ruby', '~> 1.0'
-  spec.add_dependency 'hanami-utils',    '~> 1.3'
+  spec.add_dependency 'dry-inflector',   '~> 0.2'
 
   spec.add_development_dependency 'bundler', '>= 1.6', '< 3'
   spec.add_development_dependency 'rake',  '~> 13.0'
--- a/lib/dry/cli/banner.rb
+++ b/lib/dry/cli/banner.rb
@@ -104,7 +104,7 @@ module Dry
       # rubocop:disable Metrics/AbcSize
       def self.extended_command_options(command)
         result = command.options.map do |option|
-          name = Hanami::Utils::String.dasherize(option.name)
+          name = Inflector.dasherize(option.name)
           name = if option.boolean?
                    "[no-]#{name}"
                  else
--- a/lib/dry/cli/command_registry.rb
+++ b/lib/dry/cli/command_registry.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'concurrent/hash'
-require 'hanami/utils/callbacks'
 
 module Dry
   class CLI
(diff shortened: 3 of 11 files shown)


lib/dry/cli/command_registry.rb
```diff
@@ -206,6 +205,33 @@ def after_callbacks
...
+        # @api private
+        def initialize
+          @chain = []
+        end
+
+        # @since 0.4.0
+        # @api private
+        def append(&clbk)
+          @chain.push(clbk)
+          @chain.uniq!
```
AlfonsoUceda: Don’t use an array if you need uniqueness, better use Set data structure ;)

lib/dry/cli/command_registry.rb
```diff
@@ -206,6 +205,33 @@ def after_callbacks
           @node.after_callbacks
         end
       end
+
+      # Stack for callbacks
+      #
+      # @since 0.4.0
+      # @api private
+      class CallbackStack
```
AlfonsoUceda: I wouldn’t call it Stack because it isn’t, what do you think if we call it Chain like in Hanami utils?

lib/dry/cli/banner.rb
```diff
@@ -104,7 +105,7 @@ def self.extended_command_arguments(command)
       # rubocop:disable Metrics/AbcSize
       def self.extended_command_options(command)
         result = command.options.map do |option|
-          name = Hanami::Utils::String.dasherize(option.name)
+          name = Dry::Inflector.new.dasherize(option.name)
```
solnic: It'd be better to make inflector configurable or at least as a `Dry::CLI::Inflector` constant.

lib/dry/cli/command_registry.rb
```diff
@@ -206,6 +205,33 @@ def after_callbacks
...
+        end
+
+        # @since 0.4.0
+        # @api private
+        def run(context, *args)
+          @chain.each do |callback|
+            context.instance_exec(*args, &callback)
+          end
+        end
+      end
```
solnic: Please use attr reader for `@chain`

dry-cli.gemspec
```diff
@@ -27,7 +27,7 @@ Gem::Specification.new do |spec|
   end
 
   spec.add_dependency 'concurrent-ruby', '~> 1.0'
-  spec.add_dependency 'hanami-utils',    '~> 1.3'
+  spec.add_dependency 'dry-inflector',   '~> 0.2.0'
```
solnic: I'd recommend `~> 0.2` otherwise it won't allow you to install dry-inflector `> 0.2`

PR #66, opened 2019-12-16
Anonymous registry syntax

IvanShamatov:
I would love to suggest an anonymous registry syntax. 
Usual behavior expect a developer to define a `registry`-module and pass it into Dry::CLI constructor
```ruby
module Commands
  extend Dry::CLI::Registry

  register 'version', Version, aliases: %w[v -v --version]
  register 'echo',    Echo

  register 'generate', aliases: ['g'] do |prefix|
    prefix.register 'config', Generate::Configuration
  end
end

Dry::CLI.new(Commands).call
```

IMO developer can easily skip this step with an understanding of what registry is and implementing it, just by using a block-syntax, which create a module in runtime.

```ruby
Dry::CLI.new do |cli|
  cli.register 'version', Version, aliases: %w[v -v --version]
  cli.register 'echo',    Echo

  cli.register 'generate', aliases: ['g'] do |prefix|
    prefix.register 'config', Generate::Configuration
  end
end.call
```

or even a cleaner way using @AMHOL & @solnic suggestions:

```ruby
Dry.CLI do
  register 'version', Version, aliases: %w[v -v --version]
  register 'echo',    Echo

  register 'generate', aliases: ['g'] do
    register 'config', Generate::Configuration
  end
end.call
```
It is not really a line-saver but decreases an entry threshold for a potential user.


lib/dry/cli.rb
```diff
@@ -39,8 +39,14 @@ def self.command?(command)
     #
     # @return [Dry::CLI] the new instance
     # @since 0.1.0
-    def initialize(registry)
-      @commands = registry
+    def initialize(registry = nil)
+      @commands = if block_given?
+                    anonymous_registry = Module.new { extend(Dry::CLI::Registry) }
+                    yield(anonymous_registry)
```
AMHOL: Could use an arity check here to `instance_eval` or `yield` depending on the number of block arguments and act as a DSL/injection.

solnic:
In other dry-rb gems we have a convention to use a builder method that creates an object that's already ready to be used. Maybe it's worth doing the same here, so ie:

```ruby
Dry.CLI do
   #
end
```

That would be in-line with ie dry-schema and its `Dry::Schema.Params do ... end` or dry-types' `include Dry.Types()`

WDYT?

solnic:
> The plain object syntax (Dry::CLI.new) is the most natural way that we can conceive this. The most natural thing that you can do in Ruby is to instantiate an object, I'm not convinced by Dry.CLI

I 100% agree with this BUT...please reconsider having `Dry::CLI` as a class, because `Dry::CLI` is **the namespace** too. That's one of the reasons why we have core classes called differently in other dry-rb gems. Using classes as namespaces is cumbersome in the long term and we try not to do that.

There are nice advantages with a builder method - it's more flexible than relying on a constructor, you can easily do more under the hood, you can hide many implementation details, whereas with a constructor you're **bound to it** and it's harder to change things later on.

We have `Dry::Inflector.new ...` already so having `Dry::CLI.new...` is fine but in the long term I think it would be really good to have a very consistent approach, so that all gems work in the same way when it comes to high-level DSL that builds a core object.

IvanShamatov:
I've added both syntax right now. 
What I was trying to do is not to save keystrokes or lines entered, but I wanted to reduce the number of concepts to understand when you start using this gem. In my understanding, knowing what registry is a little bit useless. That way, if a registry can be hidden from the final user, why not to do so?

@solnic It is a good point to make it consistent with other gems in Dry-rb, which I believe all devs will benefit from.

@jodosha And an awesome comparison with Hanami::Router. For example, I would not mind having syntax like that:

```ruby
cli = Dry::CLI.new
cli.register 'cmd', Command
cli.call
```
But let's keep it in mind for the future.

P.S.
I almost completely removed integration specs. I just left one spec per file to truly check it outside the system
But I've added a unit spec `spec/unit/dry/cli/cli_spec.rb` with full number of previous specs and test them against all possible notations.

PR #72, opened 2020-01-11
Fix aliases for boolean options bug

IvanShamatov:
This PR fixes aliases for options, mainly.
So what was the bug previously, where aliases mapped in to a string. 
And to explicitly say that CLI expect value after SPACE, we need to form string like that:
```
- aliases.join(' ')
+ aliases.map { |name| "-#{name} VALUE" }
```
And it fixed initial problem: https://github.com/dry-rb/dry-cli/pull/68/files

Though it is not working for boolean options where there is no value for an option. Its appearance is enough to understand the value.

So i fixed it that way:
1. aliases '-p', 'p', '--p' — are treated as the same, because they are singular letter and CLI expect -p flag. (
2. words are prefixed with `--`, you can not pass option `-port`, only `--port` (https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html)
3. banner is fixed as well

There is one more thing I found out. Combining options works perfect!

### DIFF
--- a/lib/dry/cli/banner.rb
+++ b/lib/dry/cli/banner.rb
@@ -110,8 +110,7 @@ module Dry
                  else
                    "#{name}=VALUE"
                  end
-
-          name = "#{name}, #{option.aliases.map { |a| a.start_with?('--') ? "#{a}=VALUE" : "#{a} VALUE" }.join(', ')}" unless option.aliases.empty? # rubocop:disable Metrics/LineLength
+          name = "#{name}, #{option.alias_names.join(', ')}" if option.aliases.any?
           name = "  --#{name.ljust(30)}"
           name = "#{name}\t# #{option.desc}"
           name = "#{name}, default: #{option.default.inspect}" unless option.default.nil?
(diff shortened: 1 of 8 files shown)


solnic:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity decreasing per file
==============================
+ lib/dry/cli/banner.rb  -1
         

Clones added
============
- spec/support/shared_examples/rendering.rb  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/dry-rb/dry-cli/pullRequest?prid=4803054&bid=15982539)

IvanShamatov:
@solnic Covered with specs. Because it is almost an end-to-end test, Codacity finds duplications. But they test different and quite self-explanatory when you read specs. I recommend leaving duplications as is.

solnic:
> But they test different and quite self-explanatory when you read specs. I recommend leaving duplications as is.

this is fine, I'll look into disabling duplication metrics in specs.

PR #77, opened 2020-01-24
Inline syntax for commands

IvanShamatov:
`dry/cli/inline` is not required by default and explicit requirement of this file means that it is expected of abusing global namespace with methods below

Example: ls
```ruby
#!/usr/bin/env ruby
require 'bundler/inline'
gemfile do
  gem 'dry-cli', require: 'dry/cli/inline', 
                 github: 'dry-rb/dry-cli',                   # sure you don't need these lines
                 branch: 'feature/inline-syntax-for-command' # but I leave them for testing purposes
end

desc 'List files in a directory'
argument :path, required: false, desc: '[DIR]'
option :all, aliases: ['a'], type: :boolean

run do |path: '.', **options|
  puts options.key?(:all) ? Dir.entries(path) : Dir.children(path)
end
```

```
$ chmod +x ls
$ ./ls
```

### DIFF
--- a/lib/dry/cli/inline.rb
+++ b/lib/dry/cli/inline.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+require 'backports/2.5.0/module/define_method' if RUBY_VERSION < '2.5'
+
+module Dry
+  class CLI
+    require 'dry/cli'
+    # Inline Syntax (aka DSL) to implement one-file applications
+    #
+    # `dry/cli/inline` is not required by default
+    # and explicit requirement of this file means that
+    # it is expected of abusing global namespace with
+    # methods below
+    #
+    # DSL consists of 5 methods:
+    # `desc`, `example`, `argument`, `option` 
+    # — are similar to methods from Command class
+    #
+    # `run` accepts a block to execute
+    #
+    # @example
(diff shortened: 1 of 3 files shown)


lib/dry/cli/inline.rb
```diff
@@ -0,0 +1,44 @@
...
+#   require 'bundler/inline'
+#   gemfile { gem 'dry/cli', require: 'dry/cli/inline' }
+#
+#   desc 'List files in a directory'
+#   argument :path, required: false, desc: '[DIR]'
+#   option :all, aliases: ['a'], type: :boolean
+#
+#   run do |path: '.', **options|
+#     puts options.key?(:all) ? Dir.entries(path) : Dir.children(path)
+#   end
```
solnic: I think this doc section should be right above `Inline` module definition, otherwise yard won't pick it up correctly

lib/dry/cli/inline.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+  end
+end
+
+# `dry/cli/inline` is not required by default
+# and explicit requirement of this file means that
+# it is expected of abusing global namespace with
+# methods above
+#
+# Example:
```
solnic: This should be `@example`

PR #80, opened 2020-01-24
Revert 2.3 ruby compatibility

IvanShamatov:
Backporting `match?` for files and passed arguments for commands

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,8 @@ eval_gemfile 'Gemfile.devtools'
 
 gemspec
 
+gem 'backports', '~> 3.15.0', require: false
+
 unless ENV['CI']
   gem 'byebug', require: false, platforms: :mri
   gem 'yard',   require: false
--- a/spec/unit/dry/cli/cli_spec.rb
+++ b/spec/unit/dry/cli/cli_spec.rb
@@ -69,11 +69,19 @@ RSpec.describe 'CLI' do
       )
     end
 
-    it 'with option_one' do
-      output = capture_output { cli.call(arguments: %w[first_arg --option_one=test2]) }
+    it 'with option_one', if: RUBY_VERSION < '2.4' do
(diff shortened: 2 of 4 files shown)


Gemfile
```diff
@@ -5,6 +5,7 @@ source 'https://rubygems.org'
 eval_gemfile 'Gemfile.devtools'
 
 gemspec
```
solnic: please add a new line here

spec/unit/dry/cli/cli_spec.rb
```diff
@@ -102,7 +102,7 @@
     end
 
     it 'with option_with_default alias' do
-      output = capture_output { cli.call(arguments: %w[first_arg --option_with_default=test3]) }
+      output = capture_output { cli.call(arguments: %w[first_arg --option-with-default=test3]) }
```
solnic: how is that (and the rest) related to adding backports for ruby 2.3?

IvanShamatov: https://github.com/ruby/ruby/compare/ruby_2_3...ruby_2_4#diff-e5df590ab3efa95fd6b703249c0e87c8R1389
After ruby 2.4 you able to pass options underscored and it will be dasherized.
I passed it underscored because I called my option underscored, but according to 2.3 rules, I have to pass it dasherized (which were fixed in 2.4). And for anybody, who uses dry-cli with ruby 2.3 won't have to fix their codebase.

IvanShamatov: I've added specs to check previous behaviour for ruby >= 2.4

IvanShamatov: https://github.com/ruby/ruby/blob/ruby_2_4/lib/optparse.rb#L1389
because this line appeared (and 3 more lines with similar functionality)

solnic: ohhhh I see! thank your this insight

PR #83, opened 2020-01-31
Stderror introduction to any diagnostic output

IvanShamatov:
According to Unix's best practices, all diagnostic outputs and errors should go straight to Stderr.
Banner which reacts to `-h` flag is the result of command execution, so it is a proper output and goes to output.


IvanShamatov:
I have plans to use dry-effects to inject err & out inside commands in future

IvanShamatov:
I also think that STDOUT and STDERR is a dependency and should be passed to initializer

PR #86, opened 2020-02-16
Add support for subcommands with a valid parent command

mustangostang:
Subcommands are a very powerful feature in `dry-cli`. However, there is a case when you already have a parent command and want to provide extra args/options to a subcommand.

The only possibility to this now is to use a custom first argument, which leads to a lot of messy logic if your parent command and a specific subcommand are quite different.

This PR supports simple leaf logic traversal to allow configurations like this:

```Foo::CLI::Commands.register 'root', Foo::CLI::Commands::Root
Foo::CLI::Commands.register 'root leaf', Foo::CLI::Commands::Root::Leaf```

Where `root` and `root leaf` are two independent commands, both with their own implementation/arguments. The tree will try to match `root leaf` first and, if not found, will fallback to `root` (that can further process arguments/options).

In the usage mode, this will be presented as:

`foo root [ARGUMENT] || [SUBCOMMAND]`

### DIFF
--- a/lib/dry/cli/command_registry.rb
+++ b/lib/dry/cli/command_registry.rb
@@ -37,12 +37,14 @@ module Dry
         node   = @root
         args   = []
         names  = []
+        valid_leaf = nil
         result = LookupResult.new(node, args, names, node.leaf?)
 
         arguments.each_with_index do |token, i|
           tmp = node.lookup(token)
 
           if tmp.nil?
+            (result = valid_leaf) && break if valid_leaf
             result = LookupResult.new(node, args, names, false)
             break
           elsif tmp.leaf?
@@ -50,7 +52,8 @@ module Dry
             names  = arguments[0..i]
             node   = tmp
             result = LookupResult.new(node, args, names, true)
-            break
+            valid_leaf = result
+            break unless tmp.children?
           else
             names  = arguments[0..i]
             node   = tmp
@@ -138,6 +141,11 @@ module Dry
         # @api private
         def leaf?
           !command.nil?
+        end
+
+        # @api private
+        def children?
+          !children.empty?
         end
       end
 
--- a/lib/dry/cli/usage.rb
+++ b/lib/dry/cli/usage.rb
@@ -30,7 +30,9 @@ module Dry
       def self.commands_and_arguments(result)
         max_length = 0
         ret        = commands(result).each_with_object({}) do |(name, node), memo|
-          args = if node.leaf?
+          args = if node.leaf? && node.children?
+                   arguments(node.command) + " ||#{SUBCOMMAND_BANNER}"
+                 elsif node.leaf?
                    arguments(node.command)
                  else
                    SUBCOMMAND_BANNER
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: lib/dry/cli/command_registry.rb [hunk 1]
```diff
@@ -37,20 +37,23 @@ def get(arguments)
...
         args   = []
         names  = []
+        valid_leaf = nil
         result = LookupResult.new(node, args, names, node.leaf?)
 
         arguments.each_with_index do |token, i|
           tmp = node.lookup(token)
 
           if tmp.nil?
+            (result = valid_leaf) && break if valid_leaf
```
IvanShamatov: How about adding extra if statement? 
```ruby
if tmp.nil? && valid_leaf
  result = valid_leaf
  break
elsif tmp.nil?
  ...
```

lib/dry/cli/usage.rb
```diff
@@ -30,7 +30,9 @@ def self.call(result)
       def self.commands_and_arguments(result)
         max_length = 0
         ret        = commands(result).each_with_object({}) do |(name, node), memo|
-          args = if node.leaf?
+          args = if node.leaf? && node.children?
+                   arguments(node.command) + " ||#{SUBCOMMAND_BANNER}"
```
IvanShamatov: I believe that banner, in this case, should look like that: 
```
cmd root [ARGUMENT|SUBCOMMAND]
```

spec/integration/subcommands_spec.rb
```diff
@@ -32,4 +32,82 @@
...
+        Description:
+          leaf subcommand
+
+        Options:
+          --help, -h                      	# Print this help
+      DESC
+
+      expect(output).to eq(expected)
+    end
+  end
```
IvanShamatov: Please, put most of your specs in unit specs, in shared commands.

spec/support/fixtures/foo
```diff
@@ -438,6 +456,9 @@ Foo::CLI::Commands.register 'hello',       Foo::CLI::Commands::Hello
 Foo::CLI::Commands.register 'greeting',    Foo::CLI::Commands::Greeting
 Foo::CLI::Commands.register 'sub command', Foo::CLI::Commands::Sub::Command
 
+Foo::CLI::Commands.register 'root', Foo::CLI::Commands::Root
+Foo::CLI::Commands.register 'root leaf', Foo::CLI::Commands::Root::Leaf
```
IvanShamatov: I don't believe that anyone will find using classes for namespaces is a good idea. How about `Roots` module? 
```ruby
Foo::CLI::Commands.register 'root', Foo::CLI::Commands::Root
Foo::CLI::Commands.register 'root leaf', Foo::CLI::Commands::Roots::Leaf
```

IvanShamatov:
@mustangostang Thank you very much for your contribution. I understand that sometimes it is hard to find time to finish the work, though your PR led helped us to form a vision, how it should work and look like. @unrooty fixed commends and made a PR https://github.com/dry-rb/dry-cli/pull/89. I'm closing this PR. Though, the feature you expected to have will appear soon.

PR #89, opened 2020-03-10
Add support for subcommands with a parent command.

unrooty:
## Overview
Due to inactivity in #86 I decided to open a new PR for `subcommands with a parent command` functionality.
This PR adds possibility to register and use root command with params and subcommands.

## Details
- Add root command with params and subcommands handling.
- Improve `banner`, `usage` and commands description to show normal usage with params and subcommands.
- Add integration and unit tests for implemented logic.
- Add implemented feature description to docs.

## Example usage
```ruby
#!/usr/bin/env ruby
require "bundler/setup"
require "dry/cli"

module Foo
  module CLI
    module Commands
      extend Dry::CLI::Registry

      class Account < Dry::CLI::Command
        desc 'Information about account'

        argument :format, default: "short", values: %w[long short], desc: "Output format"

        def call(**options)
          puts "Information about account in #{options.fetch(:format)} format."
        end

        class Users < Dry::CLI::Command
          desc 'Information about account users'

          def call(**_options)
            puts "Information about account users."
          end
        end
      end

      register "account", Account
      register "account users", Account::Users
    end
  end
end

```
```
$ foo account -h
Command:
  foo account

Usage:
  foo account [FORMAT] | foo account SUBCOMMAND

Description:
  Information about account

Subcommands:
  users                         	# Information about account users

Arguments:
  FORMAT              	# Output format: (long/short)

Options:
  --help, -h                      	# Print this help
```
```
$ foo account
Information about account in short format.
```

```
$ foo account long
Information about account in long format.
```

```
$ foo account users
Information about account users.
```


spec/integration/rendering_spec.rb
```diff
@@ -9,20 +9,21 @@
...
-        foo server                             # Start Foo server (only for development)
+        foo hello                                                  # Print a greeting
+        foo new PROJECT                                            # Generate a new Foo project
+        foo root-command [ARGUMENT|SUBCOMMAND]                     # Root command with arguments and subcommands
+        foo routes                                                 # Print routes
+        foo server                                                 # Start Foo server (only for development)
         foo sub [SUBCOMMAND]
         foo variadic [SUBCOMMAND]
-        foo version                            # Print Foo version
+        foo version                                                # Print Foo version
```
IvanShamatov: @jodosha this tabs makes me crazy, in a specs, can't we remove it (replace with rjust/ljust)? :(

unrooty:
@IvanShamatov I'll look into it tomorrow.

unrooty:
@IvanShamatov, I think this spaces in specs allow to check that output correct. I would forget to add ljust between subcommand's name and description without this spaces in tests .

unrooty:
@IvanShamatov, I concidered the possibility to use prefix for command with subcommands and params.

I've changed the command registration process a little:
```ruby
# file lib/dry/cli/registry.rb

def register(name, command = nil, aliases: [], &block)
  @commands.set(name, command, aliases)

  if block_given?
    prefix = Prefix.new(@commands, name, aliases)
    if block.arity.zero?
      prefix.instance_eval(&block)
    else
      yield(prefix)
    end
  end
end
```

and now it allows us to register the commands in the following ways:

```ruby
register 'root', Root do
  register 'leaf', Roots::Leaf
end
```

```ruby
register 'root', Root

register 'root' do
  register 'leaf', Roots::Leaf
end
```

```ruby
register 'root', Root

register 'root' do |prefix|
  prefix.register 'leaf', Roots::Leaf
end
```

I hope I helped you :smile: )

lib/dry/cli/parser.rb
```diff
@@ -54,13 +54,18 @@ def self.parse_required_params(command, arguments, names, parsed_options)
...
         unless all_required_params_satisfied
           parsed_required_params_values = parsed_required_params.values.compact
+          command_name = full_command_name(names)
 
-          usage = "\nUsage: \"#{full_command_name(names)} #{command.required_arguments.map(&:description_name).join(' ')}\"" # rubocop:disable Metrics/LineLength
+          if command.subcommands.empty?
+            usage = "\nUsage: \"#{full_command_name(names)} #{command.required_arguments.map(&:description_name).join(' ')}\"" # rubocop:disable Metrics/LineLength
+          else
+            usage = "\nUsage: \"#{command_name} #{command.required_arguments.map(&:description_name).join(' ')} | #{command_name} SUBCOMMAND\"" # rubocop:disable Metrics/LineLength
+          end
```
IvanShamatov:
```ruby
usage = "\nUsage: \"#{full_command_name(names)} #{command.required_arguments.map(&:description_name).join(' ')}\""

usage += " | #{command_name} SUBCOMMAND\"" if command.subcommands.any?
```

lib/dry/cli/command.rb
```diff
@@ -48,6 +49,10 @@ def self.extended(base)
         # @since 0.1.0
         # @api private
         attr_reader :options
+
+        # @since 0.6.0
```
IvanShamatov: Please mark it like that, so that when I will prepare release it will be easy to find all occurrences `# @since 0.6.x`

lib/dry/cli/command.rb
```diff
@@ -340,6 +345,10 @@ def self.optional_arguments
         arguments.reject(&:required?)
       end
 
+      def self.subcommands
```
IvanShamatov:
```ruby
# @since 0.6.x
# @api private
```

lib/dry/cli/command_registry.rb
```diff
@@ -139,6 +145,10 @@ def aliases!(aliases)
         def leaf?
           !command.nil?
         end
+
+        def children?
```
IvanShamatov:
```ruby
# @since 0.6.x
# @api private
```

spec/support/shared_examples/commands.rb
```diff
@@ -201,5 +201,93 @@
...
+          #{cmd} root-command
+
+        Usage:
+          #{cmd} root-command ROOT_COMMAND_ARGUMENT | #{cmd} root-command SUBCOMMAND
+
+        Description:
+          Root command with arguments and subcommands
+
+        Subcommands:
+          sub-command                   	# Root command sub command
```
IvanShamatov: ❤️  I was looking for that.

lib/dry/cli/parser.rb
unrooty: In your example we will have two double quotes:
```
Usage: "foo root-command ROOT_COMMAND_ARGUMENT" | foo root-command SUBCOMMAND"
```
Maybe it is better to do something like this?
```ruby
usage = "\nUsage: \"#{full_command_name(names)} #{command.required_arguments.map(&:description_name).join(' ')}" 

if command.subcommands.empty?
  usage += '"'
else
  usage += " | #{command_name} SUBCOMMAND\""
end
```

spec/support/shared_examples/commands.rb
unrooty: Glad you found it :smile:

lib/dry/cli/parser.rb
IvanShamatov: we probably can omit it at all
`Usage: foo root-command ROOT_COMMAND_ARGUMENT | foo root-command SUBCOMMAND`

IvanShamatov: or if you want to save double quotes, we can add them at the end once:
```ruby
usage = "\nUsage: \"#{full_command_name(names)} #{command.required_arguments.map(&:description_name).join(' ')}"
usage += " | #{command_name} SUBCOMMAND" if command.subcommands.any?
usage += '"'
```

unrooty:
@IvanShamatov I have resolved all your comments.

docsite/source/commands-with-subcommands-&-params.md
```diff
@@ -0,0 +1,85 @@
...
+name: dry-cli
+---
+
+There is a way to register command with arguments, options and subcommands.
+
+This helps to implement complex nested logic.
+
+Arguments and options can be defined for parent both commands and child commands.
+
+**Note:** If you call a command with an argument equal to the name of the subcommand, it will call the subcommand instead of the parent command.
```
solnic:
```suggestion
^INFO
If you call a command with an argument equal to the name of the subcommand, it will call the subcommand instead of the parent command.
^
```

docsite/source/index.html.md
```diff
@@ -10,6 +10,7 @@ sections:
   - arguments
   - options
   - variadic-arguments
+  - commands-with-subcommands-&-params
```
solnic: Please use `commands-with-subcommands-and-params` instead (same with the corresponding file name).

spec/integration/rendering_spec.rb
solnic: I'm for removing them altogether to be honest. It's just a waste of space and it can easily make certain lines not fit and wrap or not be visible in a terminal (thus require horizontal scrolling). I understand the aesthetic aspect of such alignment but it doesn't seem to be worth the effort.

jodosha: @IvanShamatov Please proceed.

IvanShamatov:
@unrooty Ok, don't worry about tabs, I will remove all of them in one extra PR.

unrooty:
@solnic I have resolved comments connected with documentation. Should I remove spaces in specs in current PR after final decision?

IvanShamatov:
I believe, we can merge

unrooty:
@IvanShamatov @solnic @jodosha Thank you for you review and for this great gem! :smile:

PR #90, opened 2020-03-16
Remove `concurrent-ruby` as runtime dependency

jodosha:
## Enhancements

We recently discussed in chat about thread-safety, code loading, and if it was worth to keep `concurrent-ruby` around.

This proposal removes `concurrent-ruby`, and still keeps the thread-safety properties. It's a win-win situation, and `dry-cli` can have zero runtime gem dependencies.

---

One note: I removed `concurrent-ruby` gem from `project.yml`, IDK how to reflect this change in `dry-cli.gemspec`.

### DIFF
--- a/lib/dry/cli/command.rb
+++ b/lib/dry/cli/command.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'forwardable'
-require 'concurrent/array'
 require 'dry/cli/option'
 
 module Dry
@@ -26,33 +25,62 @@ module Dry
           super
 
           base.class_eval do
+            @_mutex       = Mutex.new
             @description  = nil
-            @examples     = Concurrent::Array.new
-            @subcommands  = Concurrent::Array.new
-            @arguments    = Concurrent::Array.new
-            @options      = Concurrent::Array.new
+            @examples     = []
+            @arguments    = []
+            @options      = []
+            @subcommands  = []
           end
         end
 
         # @since 0.1.0
         # @api private
-        attr_reader :description
+        def description
+          @_mutex.synchronize do
+            @description
+          end
+        end
 
         # @since 0.1.0
         # @api private
-        attr_reader :examples
+        def examples
+          @_mutex.synchronize do
+            @examples
+          end
+        end
 
         # @since 0.1.0
         # @api private
-        attr_reader :arguments
+        def arguments
+          @_mutex.synchronize do
+            @arguments
+          end
+        end
 
         # @since 0.1.0
         # @api private
-        attr_reader :options
+        def options
+          @_mutex.synchronize do
+            @options
+          end
+        end
 
         # @since 0.6.x
         # @api private
-        attr_accessor :subcommands
+        def subcommands
+          @_mutex.synchronize do
+            @subcommands
+          end
+        end
+
+        # @since 0.6.x
+        # @api private
+        def subcommands=(value)
+          @_mutex.synchronize do
+            @subcommands = value
+          end
+        end
       end
 
       # Set the description of the command
@@ -72,7 +100,9 @@ module Dry
       #     end
       #   end
       def self.desc(description)
-        @description = description
+        @_mutex.synchronize do
+          @description = description
+        end
       end
 
       # Describe the usage of the command
@@ -108,7 +138,9 @@ module Dry
       #   #     foo server --port=2306         # Bind to a port
       #   #     foo server --no-code-reloading # Disable code reloading
       def self.example(*examples)
-        @examples += examples.flatten
+        @_mutex.synchronize do
+          @examples += examples.flatten
+        end
       end
 
       # Specify an argument
@@ -202,7 +234,9 @@ module Dry
       #   #   Options:
       #   #     --help, -h          # Print this help
       def self.argument(name, options = {})
-        @arguments << Argument.new(name, options)
+        @_mutex.synchronize do
+          @arguments << Argument.new(name, options)
+        end
       end
 
       # Command line option (aka optional argument)
@@ -316,13 +350,17 @@ module Dry
       #   # Options:
       #   #   --port=VALUE, -p VALUE
       def self.option(name, options = {})
-        @options << Option.new(name, options)
+        @_mutex.synchronize do
+          @options << Option.new(name, options)
+        end
       end
 
       # @since 0.1.0
       # @api private
       def self.params
-        (@arguments + @options).uniq
+        @_mutex.synchronize do
+          (@arguments + @options).uniq
+        end
       end
 
       # @since 0.1.0
--- a/lib/dry/cli/command.rb
+++ b/lib/dry/cli/command.rb
@@ -36,51 +36,27 @@ module Dry
 
         # @since 0.1.0
         # @api private
-        def description
-          @_mutex.synchronize do
-            @description
-          end
-        end
+        attr_reader :description
 
         # @since 0.1.0
         # @api private
-        def examples
-          @_mutex.synchronize do
-            @examples
-          end
-        end
+        attr_reader :examples
 
         # @since 0.1.0
         # @api private
-        def arguments
-          @_mutex.synchronize do
-            @arguments
-          end
-        end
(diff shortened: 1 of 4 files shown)


lib/dry/cli/command.rb
```diff
@@ -72,7 +100,9 @@ def self.extended(base)
       #     end
       #   end
       def self.desc(description)
-        @description = description
+        @_mutex.synchronize do
```
IvanShamatov: this is not an array or hash, we just assing a value, should we wrap it with synchronize?

lib/dry/cli/command.rb
```diff
@@ -26,33 +25,62 @@ def self.extended(base)
...
+            @options      = []
+            @subcommands  = []
           end
         end
 
         # @since 0.1.0
         # @api private
-        attr_reader :description
+        def description
+          @_mutex.synchronize do
```
IvanShamatov: what is the reason to have synchonize for getters? I thought we only need to sync and lock structure at writing/appending time.

lib/dry/cli/command.rb
```diff
@@ -316,13 +350,17 @@ def self.argument(name, options = {})
...
+          @options << Option.new(name, options)
+        end
       end
 
       # @since 0.1.0
       # @api private
       def self.params
-        (@arguments + @options).uniq
+        @_mutex.synchronize do
+          (@arguments + @options).uniq
```
AMHOL: No mutations happening here, I don't think using a mutex is necessary?

lib/dry/cli/command.rb
AMHOL: Looks like it's necessary:

```ruby
class Counter
  def self.set(n)
    @n = n
  end

  def self.get
    @n
  end
end

test = proc do
  threads = 100.times.map { |n| Thread.new { Counter.set(n.next) } }
  threads.map(&:join)
  Counter.get
end

Array.new(10, &test)
# => [81, 94, 98, 100, 100, 100, 98, 100, 97, 100]
```

lib/dry/cli/command_registry.rb
```diff
@@ -13,57 +12,62 @@ class CommandRegistry
...
-            names  = arguments[0..i]
-            node   = tmp
-            result = LookupResult.new(node, args, names, true)
-            valid_leaf = result
-            break unless tmp.children?
-          else
-            names  = arguments[0..i]
-            node   = tmp
-            result = LookupResult.new(node, args, names, node.leaf?)
+        @_mutex.synchronize do
```
AMHOL: Don't think this one is necessary, no mutation as far as I can tell?

lib/dry/cli/command.rb
flash-gordon: It's not about mutation, it's about consistent reading. Unserialized access can lead to anomalies. From the caller perspective, you anticipate `params` returns a snapshot of consistent data. When you access `@arguments` and `@options` without coordination the result may be inconsistent. Do you see the point? I may draw a diagram for better understanding.

AMHOL: Got you, so basically, the mutex needs to synchronize around the reads and writes?

lib/dry/cli/command.rb
AMHOL: ugh, yeah, I see what's happening =P

lib/dry/cli/command.rb
flash-gordon: Yes

Darigaaz:
in my opinion cleaner approach will be
as it does not pollute `Foo` class and its singleton_class
also it incapsulates all behavour in module/modules and hides mutex from `Foo`

```ruby
class SynchronizedAccess < ::Module
  def initialize
    __mutex__= nil

    super

    define_method(:initialize) do |*args|
      __mutex__ = Mutex.new

      super(*args)
    end

    # like `def self.synchronized` but as closure with __mutex__
    # defines instance methods on our instance module
    # calls original method wrapped in mutex
    singleton_class.define_method(:synchronized) do |method_name|
      define_method(method_name) do |*args|
        __mutex__.synchronize do
          super(*args)
        end
      end
    end
  end

  # to be included in class
  # to make class method `Foo.synchronized` available
  # as a closure around our instance of SynchronizedAccess
  class ClassMethods < ::Module
    def initialize(mod)
      __mod__ = mod

      super()

      define_method(:synchronized) do |method_name|
        __mod__.synchronized(method_name)
      end
    end
  end

  # @note dont like `base.singleton_class.define_method` as it defines method on Foo.singleton_class
  def prepended(base)
    base.extend ClassMethods.new(self)

    super
  end
end


class Foo
  prepend SynchronizedAccess.new

  synchronized def foo
    puts "it's safe here"
  end
end
```

lib/dry/cli/command.rb
jodosha: @IvanShamatov @flash-gordon The reason why I wrapped getters isn't for atomicity of the operation (it is atomic indeed), but because this reference is hold at the class level and it can be theoretically be altered like this:

```ruby
class Command < Dry::CLI::Command
  example "foo", "bar"
end

Command.examples.reverse!
puts Command.examples # => ["bar", "foo"]
```

Because we can't provide proper encapsulation for this, at least a lock should guarantee thread-safety.

See https://books.google.it/books?id=mzgFCAAAQBAJ&pg=PA59&lpg=PA59&dq=information+escape+thread+safety&source=bl&ots=7W9G821iaY&sig=ACfU3U0ZOeOsLvGLSLoFnfwiQwS-6xhAAw&hl=en&sa=X&ved=2ahUKEwjQhbWayaToAhXTUhUIHU-HAScQ6AEwDHoECA0QAQ#v=onepage&q=information%20escape%20thread%20safety&f=false

WDYT?

flash-gordon: @jodosha since the mutex is attached to a command instance and not to its ivars, it won't actually help. Once `.examples` returns the array can be mutated at will, even in a thread-unsafe manner. If you want to secure it there are two ways I can think of
1. `.dup` values returned by `examples`, i.e. 
  ```ruby
  def examples
    @examples.dup
  end
  ```
2. `.freeze` values every time
  ```ruby
  def example(*examples)
    @_mutex.synchronize do
      @examples = (@examples + examples.flatten(1)).freeze
    end
  end
  ```

jodosha: @flash-gordon Yeah, on a second thought, my point isn't about thread-safety, but mutability of the state.

I'll remove the mutex around those operations.

Thanks all for the discussion.

lib/dry/cli/command.rb
```diff
@@ -108,7 +138,9 @@ def self.desc(description)
       #   #     foo server --port=2306         # Bind to a port
       #   #     foo server --no-code-reloading # Disable code reloading
       def self.example(*examples)
-        @examples += examples.flatten
+        @_mutex.synchronize do
+          @examples += examples.flatten
```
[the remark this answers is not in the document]
jodosha: I changed the code to use `flatten(1)`.

jodosha:
Folks, I've removed the synchronization around getters and setters.

I also tried to limit the amount of changes that you suggested. Specifically:

  1. I changed the examples using `flatten` to `flatten(1)`, but not `concat`. If we see improvements needed here, please open another PR to address them.
  1. I din't went for the `SynchronizedAccess` module because the final implementation has only 5 places where `Mutex` is used, IMO it isn't worth the effort to DRY it. Verbosity drives simplicity, and it's fine until it's painful, but now we're really far from it.

Please let me know what do you think about it. Thanks all for the conversation! 🙌

project.yml
```diff
@@ -11,4 +11,3 @@ gemspec:
     - [rspec, "~> 3.7"]
     - [simplecov, "~> 0.17.1"]
   runtime_dependencies:
```
flash-gordon: I think it's better to remove this key or it may break

jodosha: @flash-gordon Removed. How to reflect these changes into `dry-cli.gemspec`?

flash-gordon: @jodosha it'll be updated automatically after merging

PR #92, opened 2020-03-17
Banner and Parser small refactoring

IvanShamatov:
In this PR I removed the generation of ProgName from Banner and Parser. We used it to print error messages from this classes. Instead, I pass a one time generated progname. 

And now it behaves as it was initially planned https://github.com/dry-rb/dry-cli/blob/master/lib/dry/cli/command.rb#L177-L179.

```ruby
      # $ foo generate action
      #   ERROR: "foo generate action" was called with no arguments
      #   Usage: "foo generate action APP ACTION"
```

And made `ERROR` tag consistent in a project.

### DIFF
--- a/lib/dry/cli/parser.rb
+++ b/lib/dry/cli/parser.rb
@@ -13,7 +13,7 @@ module Dry
       # @since 0.1.0
       # @api private
       #
-      def self.call(command, arguments, names)
+      def self.call(command, arguments, prog_name)
         original_arguments = arguments.dup
         parsed_options = {}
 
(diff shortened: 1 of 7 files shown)


lib/dry/cli/parser.rb
```diff
@@ -30,22 +30,16 @@ def self.call(command, arguments, names)
...
+        parse_required_params(command, arguments, prog_name, parsed_options)
       rescue ::OptionParser::ParseError
-        Result.failure("Error: \"#{names.last}\" was called with arguments \"#{original_arguments.join(' ')}\"") # rubocop:disable Metrics/LineLength
-      end
-
-      # @since 0.1.0
-      # @api private
-      def self.full_command_name(names)
-        ProgramName.call(names)
+        Result.failure("ERROR: \"#{prog_name}\" was called with arguments \"#{original_arguments.join(' ')}\"") # rubocop:disable Metrics/LineLength
```
jodosha: @IvanShamatov Never trust user input 😉 . Original `original_arguments.join(' ')` should be passed to `Shellescape.shellescape` or `Shellescape.shelljoin(original_arguments)`.

IvanShamatov: I've never heard of this before. I mean I understand the concept of injection. So I found a couple of references https://www.reddit.com/r/ruby/comments/9lr1mk/what_is_the_use_case_for_shellwordsescape/, 
https://pypi.org/project/shellescape/
And it is said that escaping is useful when you pass results to `system/exec`  , and there is nothing bad with just printing it, even if there is an `rm -rf /`

PR #93, opened 2020-03-19
Ability to register instances in Registry

IvanShamatov:
Ability to push instances of commands to registry.
Initially, all commands had `command_name` attribute inside constructor, so that you had some problems with initialization of command objects. 
I removed it couple of minor versions ago and since then `dry-cli` freed `#initializer` to everybody.

Though, it was quite useless, since `dry-cli` instanciate commands inside itself, and developer doesn't have ability to configure command even if they can implement their own initializer method without calling super.
Now you can write something like that:

```ruby
require 'sequel'
DB = Sequel.sqlite

Dry.CLI do
  register 'version', Version # <= subclass of Dry::CLI::Command, not instance
  register 'run', Runner.new(db: DB)
end
```

### DIFF
--- a/lib/dry/cli.rb
+++ b/lib/dry/cli.rb
@@ -106,16 +106,13 @@ module Dry
     # @api private
     def perform_registry(arguments)
       result = registry.get(arguments)
+      return usage(result) unless result.found?
 
-      if result.found?
-        command, args = parse(result.command, result.arguments, result.names)
+      command, args = parse(result.command, result.arguments, result.names)
 
-        result.before_callbacks.run(command, args)
-        command.call(**args)
-        result.after_callbacks.run(command, args)
-      else
-        usage(result)
-      end
+      result.before_callbacks.run(command, args)
+      command.call(**args)
+      result.after_callbacks.run(command, args)
     end
 
     # Parse arguments for a command.
@@ -135,24 +132,33 @@ module Dry
 
       result = Parser.call(command, arguments, prog_name)
 
-      if result.help?
-        out.puts Banner.call(command, prog_name)
-        exit(0)
-      end
+      return help(command, prog_name) if result.help?
 
-      if result.error?
-        err.puts(result.error)
-        exit(1)
-      end
+      return error(result) if result.error?
 
-      [command.new, result.arguments]
+      [build_command(command), result.arguments]
     end
 
-    # Prints the command usage and exit.
-    #
-    # @param result [Dry::CLI::CommandRegistry::LookupResult]
-    # @param out [IO] sta output
-    #
+    # @since x.x.x
+    # @api private
+    def build_command(command)
+      command.is_a?(Class) ? command.new : command
+    end
+
+    # @since x.x.x
+    # @api private
+    def help(command, prog_name)
+      out.puts Banner.call(command, prog_name)
+      exit(0)
+    end
+
+    # @since x.x.x
+    # @api private
+    def error(result)
+      err.puts(result.error)
+      exit(1)
+    end
+
     # @since 0.1.0
     # @api private
     def usage(result)
--- a/lib/dry/cli/command.rb
+++ b/lib/dry/cli/command.rb
@@ -50,11 +50,11 @@ module Dry
         # @api private
         attr_reader :options
 
-        # @since 0.6.x
+        # @since x.x.x
         # @api private
         attr_reader :subcommands
 
-        # @since 0.6.x
+        # @since x.x.x
         # @api private
         attr_writer :subcommands
       end
@@ -351,7 +351,7 @@ module Dry
         arguments.reject(&:required?)
       end
 
-      # @since 0.6.x
+      # @since x.x.x
       # @api private
       def self.subcommands
         subcommands
--- a/lib/dry/cli/command_registry.rb
+++ b/lib/dry/cli/command_registry.rb
@@ -26,7 +26,10 @@ module Dry
           end
 
           node.aliases!(aliases)
-          node.leaf!(command) if command
+          if command
+            node.leaf!(command)
+            node.subcommands!(command)
+          end
 
           nil
         end
@@ -127,7 +130,13 @@ module Dry
         # @api private
         def leaf!(command)
           @command = command
-          command.subcommands = children
+        end
+
+        # @since x.x.x
+        # @api private
+        def subcommands!(command)
+          command_class = command.is_a?(Class) ? command : command.class
+          command_class.subcommands = children
         end
 
         # @since 0.1.0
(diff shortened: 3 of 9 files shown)


lib/dry/cli/command.rb
```diff
@@ -55,6 +55,10 @@ def self.extended(base)
         attr_accessor :subcommands
       end
 
+      # @since 0.6.x
+      # @api private
+      attr_accessor :_context
```
solnic: @IvanShamatov I would make this `attr_reader` and in the builder method I'd do `instance_variable_set("@_context", context)`. Otherwise it *seems* like this attribute could be changed more than once which I believe is not the case.

lib/dry/cli/command_context.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+require 'ostruct'
+
+module Dry
+  class CLI
+    # CommandContext
+    #
+    # @since 0.6.x
+    class CommandContext < OpenStruct
```
solnic: Please *do not* use `OpenStruct` for this. It responds to any method and returns `nil` even when there's no corresponding attribute. There are very few cases where `OpenStruct` is a good choice and this isn't one. Even if we want to support arbitrary attributes it's still better to have an internal hash + method_missing based on `key?` checks.

IvanShamatov: This implementation of Context is based on https://github.com/hanami/hanami/blob/master/lib/hanami/cli/commands.rb#L63 internal hanami Context. I believe I can improve it a little bit. sure, will fix it (was thinking about it too)

IvanShamatov: I currently replaced with primitive implementation, just for now. Later, when I'll start adding generation abilities, it will evolve

IvanShamatov:
@jodosha Thanks for your review, I split this PR into 2. This one will end up with only having ability to pass instance to registry.

lib/dry/cli.rb
```diff
@@ -135,23 +132,50 @@ def parse(command, arguments, names)
...
+    end
 
-      [command.new, result.arguments]
+    # Prepares command instance and injects context
+    #
+    # @param command [Dry::CLI::Command]
+    #
+    # @since 0.7.0
+    # @api private
+    def command_builder(command)
```
timriley: Could we rename this to `build_command`? Feels like a more direct method name to me.

lib/dry/cli.rb
```diff
@@ -135,23 +132,50 @@ def parse(command, arguments, names)
...
+      instance = command.is_a?(Class) ? command.new : command
+      instance
```
timriley:
```suggestion
      command.is_a?(Class) ? command.new : command
```

lib/dry/cli.rb
```diff
@@ -135,23 +132,50 @@ def parse(command, arguments, names)
...
-        err.puts(result.error)
-        exit(1)
-      end
+      return error(result) if result.error?
+
+      [command_builder(command), result.arguments]
+    end
 
-      [command.new, result.arguments]
+    # Prepares command instance and injects context
```
timriley: "injects context" part is no longer relevant given you've reduced the scope of this PR?

lib/dry/cli.rb
```diff
@@ -135,23 +132,50 @@ def parse(command, arguments, names)
...
+    #
+    # @param command [Dry::CLI::Command]
```
timriley: Shouldn't this reflect that it accepts a command class _or_ instance?

More generally: do we actually _want_ full yardocs for clearly internal, private helpers? /cc @jodosha for your thoughts on this. It feels like an unnecessary amount of ceremony/maintenance burden in this case...

lib/dry/cli/command.rb
```diff
@@ -50,7 +50,7 @@ def self.extended(base)
         # @api private
         attr_reader :options
 
-        # @since 0.6.x
+        # @since 0.7.0
```
timriley: Is this and all the other `# @since` changes meant to be included? They seem unrelated to the rest of the PR...

lib/dry/cli/command_registry.rb
```diff
@@ -123,7 +123,11 @@ def lookup(token)
         # @api private
         def leaf!(command)
           @command = command
-          command.subcommands = children
+          if command.is_a?(Class)
+            command.subcommands = children
+          else
+            command.class.subcommands = children
+          end
```
timriley: This strikes me as a bit of a design smell (both the switching on class, and the mutation of the command)... I wonder if it hints at a missing abstraction that could make this nicer?

I'm OK to leave this as is for now, but perhaps it is something we should revisit soon.

lib/dry/cli.rb
IvanShamatov: this one is left like that because I will pass _context in the instance one PR later, so it will be
```ruby
instance = command.is_a?(Class) ? command.new : command
instance.set_instance_variable("@_context", Context.new...) 
instance
```

lib/dry/cli/command.rb
IvanShamatov: this is because it was submitted before in another PR

lib/dry/cli.rb
jodosha: @timriley To me, the minimum set of API docs for private helpers is `@api private` and `@since`. The first one warns users to not use that method, the second one is for our own ease of maintenance.

On Hanami, we use to set `@since x.x.x` and before the release we change all the instances of that `x.x.x` with the actual version that is about to be released.

lib/dry/cli.rb
timriley: Right. Let’s just save that change for when we do that, then? Ideally each PR/commit etc makes sense on its own. :)

lib/dry/cli/command.rb
timriley: Sorry, I still don’t understand... this PR isn’t bumping a version, and it’s not altering these methods... so why are we modifying their doc comments?

lib/dry/cli.rb
timriley: Right. Maybe we could just shrink the comments down to that minimum set? Having full “documentation” for trivial, self-explanatory methods seems unnecessary to me, and it makes the class file noisier and overall more work to maintain. What do you think?

lib/dry/cli/command.rb
IvanShamatov: Because the previous PR with these changes was merged already, it contained 0.6.x comments. `0.6.x` is easy to find in a project to bump all docs simultaneously at the moment of new release.  
@jodosha asked to fix `API docs report @since 0.6.x, but they should be @since 0.7.0.` and I just did it. These methods will first appear in 0.7.0. So this comment is 100% valid to have.

timriley: OK, makes sense, thanks for explaining!

PR #98, opened 2020-10-08
Update inline call with keyword arguments

flash-gordon:
This fixes ruby-3 build


flash-gordon:
2.3 fails for an unrelated reason, rubocop is now 2.4+

flash-gordon:
I'll merge it but I won't cut a release

PR #106, opened 2021-01-31
Adding ability to create subclasses of classes of Dry::CLI::Command

### LINKED ISSUE #105: Subclasses of commands do not inherit options and params.
## Describe the bug

When creating a subclass of a class, which inherits Dry::CLI::Command behaviour, options and arguments, that are declared in Base class are not inherited in children classes.

## To Reproduce

```ruby
class Base < Dry::CLI::Command
  option :debug

  def call(**args); puts args.inspect; end
end

class Command1 < Base
end

module Commands
  extend Dry::CLI::Registry

  register 'run', Command1
end

Dry::CLI(Commands).call

```

```
$ ./cli run --debug=test
Error: "command" was called with arguments "--debug=test"
```

## Expected behavior

```
 $ ./cli command --debug=test 
{:debug=>"test"}
```

## My environment

- Affects my production application: **NO**
- Ruby version: 2.7
- OS: mac os

IvanShamatov:
Resolves #105

Say you have a heroku-like cli, which requires to have APP_NAME argument to almost all commands, it's useful to have ability to inherit options and arguments like that:

```ruby
module Heroku
  extend Dry::CLI::Registry

  class Base < Dry::CLI::Command
    argument :app, desc: 'Application name', type: :string, required: true
    option :verbosity, desc: 'Verbosity level', type: :string, default: 'INFO'

    def call(app:, **options)
      puts "#{self.class} - App: #{app} - Options: #{options.inspect}"
    end
  end

  class Run < Base
    argument :cmd, desc: 'Command to execute', required: true

    def call(app:, cmd:, **options)
      puts "Run - App: #{app} - Command: #{cmd} - Options: #{options.inspect}"
    end
  end

  class Logs < Base
    option :num, desc: 'number of lines to display'
    option :tail, desc: 'continually stream log', type: :boolean
  end

  class Addons < Base
    option :json, desc: 'return add-ons in json format', type: :boolean, default: false
  end
  
  register 'run', Run
  register 'logs', Logs
  register 'addons', Addons
end
```

Though the limitation was set by instance_class variable, so I just dup value of this params so that they are not shared between subclasses, but behave like "inherited" if you make a subclass of a command.

```bash
$ ./heroku run MyApplicationName CommandToRun
Run - App: MyApplicationName - Command: CommandToRun - Options: {:verbosity=>"INFO"}

$ ./heroku logs MyApplicationName --tail
Based::Logs - App: MyApplicationName - Options: {:verbosity=>"INFO", :tail=>true}
```


solnic (approved):
Good stuff!

PR #108, opened 2021-02-01
Drop 2.3 ruby support and rubocop -a

IvanShamatov:
Drop 2.3 ruby support and rubocop -a for whole project


.github/workflows/ci.yml
```diff
@@ -29,7 +29,6 @@ jobs:
         - '2.5'
         - jruby
         - '2.4'
-        - '2.3'
```
solnic, line 32: this won't work, you need to add `.github/workflows/custom.ci` and include just this node there, it'll be merged into the final ci config via devtools GH action.

solnic (changes requested):

.github/workflows/ci.yml
IvanShamatov: fixed, but that way it makes current GH actions for 2.3 fail

solnic (approved):

.github/workflows/ci.yml
solnic: @IvanShamatov I'll remove 2.3 from required builds

PR #109, opened 2021-02-05
103 Safely rescue interrupts

### LINKED ISSUE #103: Safely rescue Interrupt
If a ruby program is interrupted by `Ctrl^C` or sent a `SIGINT` signal, the `Interrupt` exception is raised, causing an ugly backtrace to be printed. Ideally, `Dry::CLI` should `rescue Interrupt` and exit with a status of 130 (see: https://tldp.org/LDP/abs/html/exitcodes.html).

IvanShamatov:
Resolves #103 
 

> If a ruby program is interrupted by Ctrl^C or sent a SIGINT signal, the Interrupt exception is raised, causing an ugly backtrace to be printed. Ideally, Dry::CLI should rescue Interrupt and exit with a status of 130 (see: https://tldp.org/LDP/abs/html/exitcodes.html).

Now works as expected

### DIFF
--- a/spec/integration/signal_trap_spec.rb
+++ b/spec/integration/signal_trap_spec.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+RSpec.describe "SignalException handler" do
+  it "returns correct statuscode" do
+    pid1, pid2, pid15 = Array.new(3).map { Process.spawn("foo generate app SomeAppName") }
+    sleep 0.5 # time to start app
+
+    Process.kill("HUP", pid1)
+    Process.kill("INT", pid2)
+    Process.kill("TERM", pid15)
+
+    _, status1 = Process.wait2(pid1)
+    _, status2 = Process.wait2(pid2)
+    _, status15 = Process.wait2(pid15)
+
+    expect(status1.exitstatus).to eq(129)
+    expect(status2.exitstatus).to eq(130)
+    expect(status15.exitstatus).to eq(143)
+  end
+end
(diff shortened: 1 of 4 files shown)


spec/integration/signal_trap_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+RSpec.describe "SignalException handler" do
+  it "returns correct statuscode" do
+    pid1, pid2, pid15 = Array.new(3).map { Process.spawn("foo generate app SomeAppName") }
+    sleep 0.5 # time to start app
+
```
IvanShamatov, line 7: Groupped into one spec to spend less time in integration specs

spec/unit/dry/cli/cli_spec.rb
```diff
@@ -120,4 +120,5 @@
       )
     end
   end
+
```
solnic, line 123: Extra empty line here

lib/dry/cli/command.rb
```diff
@@ -13,37 +13,20 @@ class Command
...
-
-            @options =
-              if base.superclass.instance_variable_defined?(:@options)
-                base.superclass.instance_variable_get(:@options).dup
-              else
-                []
-              end
-          end
-        end
-
```
solnic, line 46: This seems like an unrelated change or am I missing something?

IvanShamatov: Yep, @jodosha asked to make this nicer after I merged previous PR

solnic: @IvanShamatov OK in the future please use separate PRs with such unrelated improvements. Our future selves will appreciate it and it's a good habit to have in general. It makes maintenance simpler, especially when it turns out something needs to be reverted etc.

IvanShamatov: will do that in separate pr

solnic (approved):

PR #110, opened 2021-02-07
Refactoring: Make class inheritable attributes cleaner

IvanShamatov:
As it was asked before, subj

### DIFF
--- a/lib/dry/cli/command.rb
+++ b/lib/dry/cli/command.rb
@@ -13,37 +13,20 @@ module Dry
       # @api private
       def self.inherited(base)
         super
+        base.class_eval do
+          @_mutex       = Mutex.new
+          @description  = nil
+          @examples     = []
+          @subcommands  = []
+          @arguments = base.superclass_arguments || []
+          @options = base.superclass_options || []
+        end
         base.extend ClassMethods
       end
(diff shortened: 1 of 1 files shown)


lib/dry/cli/command.rb
```diff
@@ -13,37 +13,20 @@ class Command
       # @api private
       def self.inherited(base)
         super
+        base.class_eval do
+          @_mutex       = Mutex.new
+          @description  = nil
+          @examples     = []
+          @subcommands  = []
+          @arguments = base.superclass_variable_dup(:@arguments) || []
```
jodosha, line 21: @IvanShamatov IMO `superclass_variable_dup` is good as a private method to not be used directly outside of the class.

I would go for

```ruby
@arguments = base.superclass_arguments || []
```

```ruby
def self.superclass_arguments
  superclass_variable_dup(:@arguments)
end
```

jodosha (commented):
@IvanShamatov I would make things even more cleaner, WDYT?

lib/dry/cli/command.rb
IvanShamatov: Ah, nice, will do that

PR #112, opened 2021-02-10
Transform `Dry::CLI::Utils::Files` into a class

jodosha:
## Enhancement

Transform `Dry::CLI::Utils::Files` from a module into a class to instantiate.
Being an object, it can leverage Dependency Injection to abstract the low level File System implementation.

This is extremely useful to unit test CLIs when they have to deal with file operations.

### Before

```ruby
require "dry/cli"
require "dry/cli/utils/files"

module Foo
  class CLI < Dry::CLI::Command
    def initialize(files: Dry::CLI::Utils::Files)
      @files = files
    end
  end
end
```

### After

**Notice the `.new` addition.**

```ruby
require "dry/cli"
require "dry/cli/utils/files"

module Foo
  class CLI < Dry::CLI::Command
    def initialize(files: Dry::CLI::Utils::Files.new)
      @files = files
    end
  end
end
```

## Implementation changes

  * Transformed `Dry::CLI::Utils::Files` from a module into a class to instantiate.
  * Moved all the methods from class level (module function) to instance level.
  * Introduced `Dry::CLI::Utils::Files::FileSystem` as an abstraction for low level file operations. The implementation leverages `File` and `FileUtils` from Ruby code and stdlib.
  * `Dry::CLI::Utils::Files#initialize` accepts a dependency via keyword arg `file_system`. The default value is an instance of `Dry::CLI::Utils::Files::FileSystem`
  * `Dry::CLI::Utils::Files` can utilise Dependency Injection to provide their own file system implementation, in case they want to unit test CLIs
  *  `spec/unit/dry/cli/utils/files_spec.rb` → `spec/integration/dry/cli/utils/files_spec.rb` as the specs are working with real file system operations.
  * Code and specs are now respecting cross platform line breakage (`"\n"`)
  * Improved documentation


lib/dry/cli/utils/files/file_system.rb
```diff
@@ -0,0 +1,190 @@
...
+        class FileSystem
+          # Creates a new instance
+          #
+          # @param file [Class]
+          # @param file_utils [Class]
+          #
+          # @return [Dry::CLI::Utils::Files::FileSystem]
+          #
+          # @since x.x.x
+          def initialize(file: File, file_utils: FileUtils)
```
solnic, line 22: I'm wondering if this is going to be leveraged in any way? Seems like this class could be just a simple wrapper around File and FileUtils + a couple of additional methods.

lib/dry/cli/utils/files/file_system.rb
```diff
@@ -0,0 +1,190 @@
...
+
+          private
+
+          # @since x.x.x
+          # @api private
+          attr_reader :file
+
+          # @since x.x.x
+          # @api private
+          attr_reader :file_utils
```
solnic, line 185: No biggie but in general in dry/rom we're defining readers at the top and we never-ever define them as private, because it makes debugging horrible and also method visibility can always be bypassed in Ruby, and people do that, so we can mark things as private but that won't prevent misusage. In general my belief is that documenting API as private is enough. Another nice benefit is that when you change your mind, it's just a documentation update, no code changes needed.

I *know* people have completely opposite opinions though so it'd be good to find consensus here. Again, no biggie but let's address this at some point 🙂

lib/dry/cli/utils/files/file_system.rb
jodosha: @solnic I didn't get you here. Do you mean that:

  1. `FileSystem` doesn't make sense to exist?
  2. Nobody will leverage DI in `#initialize` and so we can hardcode deps?
  3. `FileSystem` should expose class method interface?

lib/dry/cli/utils/files/file_system.rb
jodosha: @solnic I don't agree here, but consistency across libs is top priority. Fixed in https://github.com/dry-rb/dry-cli/pull/112/commits/14be7d3f9466bde94458bc16bd600b39b220f568

lib/dry/cli/utils/files/file_system.rb
```diff
@@ -0,0 +1,188 @@
...
+          #   require "dry/cli/utils/files/file_system"
+          #
+          #   fs = Dry::CLI::Utils::Files::FileSystem.new
+          #   fs.mkdir("/usr/var/project/file.rb")
+          #   # creates all the directory structure (/usr/var/project)
+          #   # where file.rb will eventually live
+          #
+          # @since x.x.x
+          # @api private
+          def mkdir_p(path, **kwargs)
```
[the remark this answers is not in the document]
IvanShamatov: How about creating `touch_p` or extend current `touch` functionality to create directory structure and then create file there? IMO it will be more clear than the difference between mkdir (which uses mkdir_p of FileUtils) and mkdir_p of FileSystem here

IvanShamatov (commented):
Don't you think that having ability to require both FileSystem and Files, will make people misuse it, or not use it at all since the idea of having wrapper on wrapper on FileUtils is not that obvious?

lib/dry/cli/utils/files/file_system.rb
solnic: @jodosha no, it makes a lot of sense to have this class - but I did mean that maybe it doesn't make sense to inject `File` and `FileUtils` and just use them internally as an implementation detail w/o providing the ability to override them. It could become a class-only interface too, actually.

lib/dry/cli/utils/files/file_system.rb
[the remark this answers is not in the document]
jodosha: @IvanShamatov it doesn't create a file. You pass the path that includes the file, but it ignores it.

```ruby
fs.mkdir("/usr/var/project/file.rb") # => creates /usr/var/project
```

IvanShamatov (approved):

jodosha:
Extracted into `dry-files`

PR #116, opened 2021-05-25
Fix documentation for Command.option

rkh:
Examples were using param instead of option.


solnic (approved):

solnic:
Thank you 🙂 Super nice to see you here 🙂

PR #120, opened 2022-04-01
Feature: free form `--help` text

jodosha:


rngtng:
oh nice, I was looking exactly for this.. what's the status?

lib/dry/cli/command.rb
```diff
@@ -108,6 +109,46 @@ def self.example(*examples)
...
+      # @param blk [Proc] a block that must return a free form string to add
+      #   extra information
+      #
+      # @since 0.8.0
+      #
+      # @example
+      #   require "dry/cli"
+      #
+      #   class Convert < Dry::CLI::Command
+      #     help do
```
rngtng, line 123: Have you though to support just `help "String"` too?

jodosha: @rngtng Yeah, that could work too.

jodosha:
@rngtng The project where I'm using this feature is low priority. I want this feature to be driven by real world needs, that's why this PR is not finished yet.

PR #122, opened 2022-10-13
Implement option required: true

capripot:
```ruby
option :option_name, desc: "This is an option", required: true
```

`required: true` is accepted and implemented by `Option`, but not used by `Parser` or `Banner`. This PR implements the enforcement of it and surface it to the help banner.

  - Adds "# REQUIRED" to option description when required
  - Check is required options is passed
  - Refactor Usage and Banner option to reuse in short usage when erroring

Checks:

  - [x] Tests are passing
  - [x] Rubocop is passing on changes contained in this PR

### DIFF
--- a/lib/dry/cli/banner.rb
+++ b/lib/dry/cli/banner.rb
@@ -135,26 +135,37 @@ module Dry
         end.join("\n")
       end
 
+      # @since 1.5.0
+      # @api private
+      def self.simple_option(option)
+        name = Inflector.dasherize(option.name)
+        name = if option.boolean?
+                 "[no-]#{name}"
+               elsif option.flag?
+                 name
+               elsif option.array?
+                 "#{name}=VALUE1,VALUE2,.."
+               else
+                 "#{name}=VALUE"
+               end
+        name = "#{name}, #{option.alias_names.join(", ")}" if option.aliases.any?
+        "--#{name}"
+      end
+
+      # @since 1.5.0
+      # @api private
+      def self.extended_option(option)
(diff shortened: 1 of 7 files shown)


IvanShamatov:
I would suggest to remove "required" flag from Parser and Banner, rather then having required option. If option is required, that means, that it is not an option, but argument.

capripot:
I agree, I was also by having to write `option :flag, required: false` but wouldn't it be a breaking change? 😕

capripot:
From @IvanShamatov 
> If option is required, that means, that it is not an option, but argument.

Could we imagine options as keyword arguments tho? Sometime it makes more sense to name the argument rather than having nameless positional arguments. And keyword arguments can be required or not, but they used to be an "options Hash".

I would leave the option to the developer to choose the style they like the best.
What do you think?

IvanShamatov:
@capripot Sure, I understand the concept of keyword arguments, I just can't remember any cli utility with that kind of required option in a wild. Do you have any examples? 

Anyway, I don't see why we can't have it. I will talk to the team to ask their opinion on that.
Thank you for the contribution

capripot:
@IvanShamatov Do you think we could merge this? ☺️ 🙏

timriley:
Hi @capripot, thank you so much for your patience here. I've started looking at this, as you might notice from the few small commits I've pushed. I'll look to get it ready to merge over the coming week, and will let you know if I have any questions for you.

cllns:
I also think the idea of a required option is a bit odd. Not really sure what else to call it though, since we already use "argument" in this project. Maybe just `required`?

PR #125, opened 2023-03-16
Add `aliases` to options

adrianthedev:
They were missing from the docs.


timriley:
Thanks @adrianthedev!

PR #127, opened 2023-06-30
doc: fix filename

noraj:
Fix the name from `filename.md` to `filename.html.md` as for other documentation pages else it causes the page to force download instead of being rendered: https://dry-rb.org/gems/dry-cli/1.0/commands-with-subcommands-and-params.


timriley:
Thanks @noraj!

PR #130, opened 2023-07-26
FIX: Wrong extension in docsite file

AngelGuerra:
Hello folks! I have seen that there is a problem in one of the links in the documentation:

If you navigate to `https://dry-rb.org/gems/dry-cli/main/` and click in the menu on `Commands with subcommands and params`, it will result in downloading a file instead of navigating. I suspect it is because the file extension is `.md` instead of `.html.md`, this PR fixes that error.

Thanks for your work!

## How to reproduce

- Go to [`https://dry-rb.org/gems/dry-cli/main/`](https://dry-rb.org/gems/dry-cli/main/).
- Click on `Commands with subcommands and params`.

Or...

- Go to [`https://dry-rb.org/gems/dry-cli/main/commands-with-subcommands-and-params`](https://dry-rb.org/gems/dry-cli/main/commands-with-subcommands-and-params).


timriley:
Fixed by #127. Thanks for submitting this!

PR #134, opened 2024-07-10
Fix CI for Ruby 3.3

cllns:
This has been broken for a while now. `Struct.new` (with no args) used to raise an error in Ruby 3.2 and earlier, but in 3.3 it's allowed. So in this PR we make a new class that actually requires an arg so the test raises the expected error


timriley:
Thanks @cllns! Pulled this right in.

PR #135, opened 2024-07-24
Support root commands that doesn't implement `call`

### LINKED ISSUE #96: Allow adding subcommand toplevel documentation (usage)
It would be nice to be able to add documentation to command *prefixes*. The way `dry-cli` deals with subcommands disallows adding documentation to be shown calling `--help` over the subcomand *prefix*.

## Examples

Given the code on this example:

```
require "dry/cli"

module Foo
  module Commands
    extend Dry::CLI::Registry

    module Generate
      class App < Dry::CLI::Command
      end

      class Action < Dry::CLI::Command
      end
    end

    register "generate", aliases: ["g"] do |prefix|
      prefix.register "app",    Generate::App
      prefix.register "action", Generate::Action
    end
  end
end
```

It would be nice to be able to add documentation over the 'generate' subcommand. So its help will be shown like:

```
Description:

here you can include a general documentation about the subcommand.

Commands:
  foo.rb generate action
  foo.rb generate app

```

A possible implementation could be:

```
    register "generate", aliases: ["g"] do |prefix|
      prefix.desc 'HERE YOU CAN INCLUDE THE DOCUMENTATION OF THE COMMAND'
      prefix.register "app",    Generate::App
      prefix.register "action", Generate::Action
    end

```

I checked the code and saw that `Prefix` class was just syntax sugar for `self.register`, so no idea on how this could be implemented, or whether there is already a way of achieving what I'm looking for.

Thanks in advance! and thanks for `dry-cli`!

gustavothecoder:
Resolve #96

I would like to add descriptions to my root commands without implementing a proper command, like a namespace.

### Example
**Command**
```ruby
class Namespace < Dry::CLI::Namespace
  desc "This is a namespace"

  class MyCommand < Dry::CLI::Command
    desc "I'm a concrete command"

    def call(**params)
      puts "I'm a concrete command"
    end
  end
end
```

**Result**

image

### DIFF
--- a/spec/support/fixtures/shared_commands.rb
+++ b/spec/support/fixtures/shared_commands.rb
@@ -440,7 +440,7 @@ module Commands
     end
   end
 
-  class Namespace < Dry::CLI::Command
+  class Namespace < Dry::CLI::Namespace
     desc "This is a namespace"
 
     class SubCommand < Dry::CLI::Command
--- a/spec/support/fixtures/shared_commands.rb
+++ b/spec/support/fixtures/shared_commands.rb
@@ -440,6 +440,18 @@ module Commands
     end
   end
 
+  class Namespace < Dry::CLI::Command
+    desc "This is a namespace"
+
+    class SubCommand < Dry::CLI::Command
+      desc "I'm a concrete command"
+
+      def call(**params)
+        puts "I'm a concrete command"
+      end
+    end
+  end
+
   class InitializedCommand < Dry::CLI::Command
     attr_reader :prop
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 
 ### Added
 
+- Support for command namespaces (@gustavothecoder in #135)
+
 ### Changed
 
 ### Deprecated
--- a/lib/dry/cli.rb
+++ b/lib/dry/cli.rb
@@ -116,6 +116,7 @@ module Dry
       return spell_checker(result, arguments) unless result.found?
 
       command, args = parse(result.command, result.arguments, result.names)
+      return usage(result) unless command.respond_to?(:call)
 
       command.instance_variable_set(:@err, err) unless command.instance_variable_defined?(:@err)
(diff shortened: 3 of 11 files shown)


spec/support/fixtures/shared_commands.rb
```diff
@@ -438,6 +438,18 @@ def call(**params)
     end
   end
 
+  class Namespace < Dry::CLI::Command
```
benoittgt, line 441: Isn't bizarre to have a "command" that the uniq usage is to print top level usage? I am wondering if it should be a better idea to move the usage of top class as specific dsl. Like suggested here https://github.com/dry-rb/dry-cli/issues/96 ?

benoittgt:
This is a great proposal. I would love to be able to use this in our CLI.

spec/support/fixtures/shared_commands.rb
gustavothecoder: It is. I tried to implement the suggestion from #96, but I couldn't figure out how to do it. Maybe we can create the `Dry::CLI::Namespace` class, so It will look less strange.

benoittgt: That's a good idea.

timriley: @gustavothecoder I like the idea of this being a specialised `Namespace` class. That way, when the class is used, it will be clearer that its purpose is to provide information about a grouping of subcommands only, and not its own distinct command behaviour.

I prefer this help text residing in a concrete class instead of being part of the top-level command registration API, since that feels more consistent with how the rest of the CLI construction is done: a class for each thing.

If you're still interested in working on this PR, I'd love it if you could make this adjustment! Let me know what you think.

gustavothecoder: @timriley Done! I think it looks better using the `Namespace` class.

timriley:
Hi @gustavothecoder!! Happy new year :) Firstly, I want to apologise for the delay in attending to all your excellent PRs. Good news, though — in January we'll be focusing on getting a new dry-cli shipped with all your work incorporated. We'll be back in touch in the PRs. Thank you for your contributions!

PR #140, opened 2024-12-02
Add functions to style text (colors, weight, etc)

### LINKED ISSUE #121: Feature request: colorized output
This is a feature I started working on long ago for Hanami: https://github.com/hanami/hanami/pull/906

There could be two parts to this:
- Colorizing the output of `dry-cli`, like each command's `help` and the list of subcommands
- Adding something like `say` that users of the library can use to print in color (instead of using `print`).

e.g.
```ruby
say "create", color: :red
```

This should probably be like `print` rather than `puts` (no trailing newline), since people should be able to print in several colors on one line.

The concrete use-case for this could be like hanami/hanami#906, being able to colorize the output of file generation for Hanami.

Is this something we'd be interested in adding to `dry-cli`? Any thoughts on the API?

gustavothecoder:
Inspired by the issue #121, I added a module with functions to change text style. The doc `docsite/source/styling-your-output.html.md` demonstrate how it works. I used the `thor` [implementation](https://github.com/rails/thor/blob/3178667e1727504bf4fb693bf4ac74a5ca6c691e/lib/thor/shell/color.rb#L9) as a basis,  but I changed the API to make combinations shorter:
```ruby
# Thor API
set_color "Hi!", :red, :on_white, :bold
# Implemented API
red(on_white(bold("Hi!")))
``` 

Resolve #121


gustavothecoder:
Maybe this API is a better idea because it looks more like Ruby:
```ruby
stylize("Hi!").red.on_white.bold
```
In this case, `stylize` will return an object with methods to style the text.

gustavothecoder:
> > Maybe this API is a better idea because it looks more like Ruby:
> > ```ruby
> > stylize("Hi!").red.on_white.bold
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In this case, `stylize` will return an object with methods to style the text.
> 
> First of all, thanks for doing this!
> 
> When I first read this PR, I also thought of this API :)
> 
> Want to go for it?
> 
> Also, can you extract some constants like `BLUE = 34` and refactor from based on that?

Yeah, I'll let you know when I'm done refactoring.

gustavothecoder:
Hey @cllns, I'm done refactoring and I think that the solution is better now, what you think?

image

lib/dry/cli/styles.rb
```diff
@@ -0,0 +1,187 @@
...
+          chainable_update!("\e[#{ON_CYAN}m#{@text}")
+        end
+
+        # since 1.3.0
+        def on_white
+          chainable_update!("\e[#{ON_WHITE}m#{@text}")
+        end
+
+        private
+
```
cllns, line 174: Can we use an attr_reader throughout, instead of the instance variable?

```suggestion
       attr_reader :text
```

Then change almost all `@text`s to just `text`

lib/dry/cli/styles.rb
```diff
@@ -0,0 +1,187 @@
...
+        end
+
+        # since 1.3.0
+        def on_cyan
+          chainable_update!("\e[#{ON_CYAN}m#{@text}")
+        end
+
+        # since 1.3.0
+        def on_white
+          chainable_update!("\e[#{ON_WHITE}m#{@text}")
```
cllns, line 170: Can we extract a helper method for escaping like this? https://en.wikipedia.org/wiki/ANSI_escape_code#SGR

e.g.

```ruby

def select_graphic_rendition(code)
  "\e[#{code}m"
end
```

lib/dry/cli/styles.rb
```diff
@@ -0,0 +1,187 @@
...
+        end
+
+        private
+
+        # @since 1.3.0
+        # @api private
+        def chainable_update!(new_text)
```
cllns, line 177: Instead of checking and only adding the `RESET` if one doesn't exist, could we just throw the `RESET` on at the end when it's coerced to a string in `to_s`? I think that should be fine. 

In that case, stylize(stylize(...)) might get two resets at the end, but that's a no-op anyway? We could also check in there and only add it if it doesn't already end with one.

cllns (changes requested):
Thanks for that refactoring! In reviewing it, I found a few more we can do as well

gustavothecoder:
> Thanks for that refactoring! In reviewing it, I found a few more we can do as well

Thanks for the feedback! Can you review it again, please?

cllns:
Sweet, thanks!

Can you try to use recursion instead of mutation? In dry-rb, we prefer to leverage immutability, as we find it easier to reason about.

To spell it out, calling `style("hello")` would return a `Dry::CLI::Styles::StyledText` instance, then each call will wrap that in another `Dry::CLI::Styles::StyledText` instance, and so on. Then when `.to_s` is called, it adds the shell escape code and calls `.to_s` on its contents.

The `StyledText` can also hold the @escape_code. For the first call to `style`, that value will be `nil` and we can use that as a convention to throw the clear/reset at the end.

Does that make sense?

gustavothecoder:
> Sweet, thanks!
> 
> Can you try to use recursion instead of mutation? In dry-rb, we prefer to leverage immutability, as we find it easier to reason about.
> 
> To spell it out, calling `style("hello")` would return a `Dry::CLI::Styles::StyledText` instance, then each call will wrap that in another `Dry::CLI::Styles::StyledText` instance, and so on. Then when `.to_s` is called, it adds the shell escape code and calls `.to_s` on its contents.
> 
> The `StyledText` can also hold the @escape_code. For the first call to `style`, that value will be `nil` and we can use that as a convention to throw the clear/reset at the end.
> 
> Does that make sense?

It does, the code is simpler now:
```ruby
def chainable_update!(style, new_text)
  StyledText.new(
    select_graphic_rendition(style) + new_text,
    select_graphic_rendition(RESET)
  )
end
```

timriley:
I'd like to explore a few other possible API approaches before going ahead with this feature. If anyone has some thoughts, please feel free to drop them here! Otherwise, I do plan to come back in the next week or so with a suggestion.

alassek:
Some prior art here that we should consider: [pastel](https://github.com/piotrmurach/pastel)

In particular, Pastel's API allows you to pre-define reusable styles:

```ruby
error    = pastel.red.bold.detach
warning  = pastel.yellow.detach

puts error.("Error!")
puts warning.("Warning")
```

I think this makes much more sense to me, rather than constructing a class with the message text as part of the state. The style state should be long-lived and the message should pass through it as a parameter.

The current approach [confuses dependencies and arguments](https://gist.github.com/alassek/1a185a8cdae20ddd991a8cd3cd90d347#1-the-distinction-between-dependencies-and-arguments).

timriley:
@alassek this is exactly what I'd come up with in an alternative branch I was aiming to push later this week :)

PR #142, opened 2024-12-14
Fix arguments with `values` allowing any value

### LINKED ISSUE #126: Unlisted value fails for option, but passes for argument
## Describe the bug

Given the following `example`:

```ruby
#!/usr/bin/env ruby
require "bundler/setup"
require "dry/cli"

module Foo
  module CLI
    module Commands
      extend Dry::CLI::Registry

      class Start < Dry::CLI::Command
        argument :arg, values: %w(one two)
        option :opt, values: %w(one two)

        def call(arg: nil, opt: nil, **)
          puts "ARG: #{arg.inspect} / OPT: #{opt.inspect}"
        end
      end

      register "start", Start
    end
  end
end

```

The following somewhat unexpected behaviour occurs:

```
example start --opt=one      # ARG: nil / OPT: "one"
example start --opt=three    # ERROR: was called with arguments "--opt=three"
example start one            # ARG: "one" / OPT: nil
example start three          # ARG: "three" / OPT: nil     <-- OOPS
```

The last one should also result in an error IMO.

## My environment

- Affects my production application: NO (but soon :-)
- Ruby version: 3.2
- OS: macOS Ventura

gustavothecoder:
Resolve #126

As reported by the issue above, it's possible to define possible values for arguments, but we don't validate if the received value can be used. With this PR the behavior will be the same of options:

https://github.com/user-attachments/assets/c2ac885c-4c90-4dcc-9e4d-fb168d247555

### DIFF
--- a/lib/dry/cli/errors.rb
+++ b/lib/dry/cli/errors.rb
@@ -9,6 +9,10 @@ module Dry
     class Error < StandardError
     end
 
+    # @since 1.3.0
+    class UnknownValueError < Error
+    end
+
     # @since 0.2.1
     class UnknownCommandError < Error
       # @since 0.2.1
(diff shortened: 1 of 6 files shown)


timriley:
Thanks @gustavothecoder, this does the job well! I hopped into the branch and made a few small adjustments before merging, you can see them in the commits above :)

PR #143, opened 2024-12-16
Prevent duplicate arguments/options

### LINKED ISSUE #139: Prevent duplicate option names
## Describe the bug

dry-cli allows one to define duplicate options that override each other. However, the all show up when running `--help`.

While in a single class this is easy to avoid, when building many commands and sharing behavior, it's easy to accidentally include extra options. See: https://github.com/hanami/cli/pull/278 

## To Reproduce

In `test.rb`:

```
require "dry/cli"

module Foo
  module CLI
    module Commands
      extend Dry::CLI::Registry

      class Request < Dry::CLI::Command
        option :mode, default: "http", values: %w[http http2]
        option :mode, default: "foo", values: %w[bar baz]

        def call(**options) = puts "Performing a request (mode: #{options.fetch(:mode)})"
      end

      register "request", Request
    end
  end
end

```

Then
```ruby
# ruby test.rb request --help
Command:
  test.rb request

Usage:
  test.rb request

Options:
  --mode=VALUE                      # : (http/http2), default: "http"
  --mode=VALUE                      # : (bar/baz), default: "foo"
  --help, -h                        # Print this help
```

It shows both `--mode`s. Same thing happens if they're identical, too.

## Expected behavior

I would expect either:
1) The second option to override the first one and completely remove the first one from `--help` (since this is what actually happens when you try to execute it with that option. only bar/baz are valid, not http/http2.
2) An error to be raised, saying it's not allowed

## My environment

Ruby 3.3, dry-cli 1.2.0 (latest as of writing this)

gustavothecoder:
Resolve #139 

This PR resolves the issue above by always considering the last defined argument/option (i.e. the last override the previous definition).

### DIFF
--- a/lib/dry/cli/command.rb
+++ b/lib/dry/cli/command.rb
@@ -195,7 +195,12 @@ module Dry
       #   #   Options:
       #   #     --help, -h          # Print this help
       def self.argument(name, options = {})
-        @arguments << Argument.new(name, options)
+        new_arg = Argument.new(name, options)
+
+        duplicate_index = @arguments.find_index { _1.name == new_arg.name }
+        @arguments.delete_at(duplicate_index) unless duplicate_index.nil?
+
+        @arguments << new_arg
       end
 
       # Command line option (aka optional argument)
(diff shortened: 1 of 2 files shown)


lib/dry/cli/command.rb
```diff
@@ -195,7 +195,12 @@ def self.example(*examples)
       #   #   Options:
       #   #     --help, -h          # Print this help
       def self.argument(name, options = {})
-        @arguments << Argument.new(name, options)
+        new_arg = Argument.new(name, options)
+
+        duplicate_index = @arguments.find_index { _1.name == new_arg.name }
+        @arguments.delete_at(duplicate_index) unless duplicate_index.nil?
```
timriley: Thanks @gustavothecoder! 

To make this simpler (and faster), I think we could convert `@arguments` and `@options` to be a hash, so that we simply replace them as a new one is registered, rather than seeking through the array. I'm going to merge this as-is, though, and open an issue for this small optimisation.

Issue #129, opened 2023-07-20
More powerful options


noraj:
## Flags

[options](https://dry-rb.org/gems/dry-cli/1.0/options/) are always considered having a value `--mode=http2` or `--mode http2`.

- [ ] Having flags (options without value)

E.g. just having `--color`, having the flag equals `options.fetch(:color)` returning true, not having the flag equals `options.fetch(:color)` returning false. It's better than having `--color true` and `--color false` with `values: %w[true false]` and casting the string to a boolean.

## Aliases

It would be nice to have has many aliases as we want for an option.

- [ ] Having aliases

E.g. `-c, --color` or `--color, --colour`.

It's often handy to have long flags that are mnemotechnic and short flags for those familiar with the tool that want to type faster.

## Mutually exclusive

Sometimes it is required to have mutually exclusive options, see #128.

- [ ] Having mutually exclusive options #128

## Casting type

As passed from the CLI, all values are strings, so it would be handy to have cast types for options.

- [ ] Having cast types

E.g. `--age 21` or `--advanced true`. It's allows to cast type in the definition of the option, avoiding forgetting to cast it when the value is retrieved or to cast it several times if the option is used at multiple places.

MadBomber:
I think a few of these have already been put in place.  For example I'm using ...
```ruby
option :debug,
  aliases:  %w[ -d --debug],
  type:  :boolean,
  default:  false
```

amd the same structure for `verbose` in a `Base` command which defines all blobal options for all toerh commands.

MadBomber:
One of the things that I found myself wishing for is some kind of `action` that gets executed when the option is present.  As is, those kinds of things have to be build as a command.   I'd much rather put a simple Proc or Lambda in the option definition that have to layout a small command class.  For example consider a `version` option.

```ruby
option :version,
  aliases:  %w[ --version ],
  action: -> { puts VERSION; exit(0) }
```
... of course you'd going to ask where does that version constant come from?  I'm assuming it comes from the command class.  If I wanted it to come from the ProgramName I would have coded it with that name space.

As is, I had to create a Version command class and then for all my commands which have versions I had to register both the command and the command with a version sub-command.

```ruby
register "my_command", self
register "my_command version', Command::PrintVersion.new(VERSION), aliases: %w[-v --version]
```

I've been experimenting in https://github.com/MadBomber/experiments/tree/master/cli_options/dry-cli

[...]

PR #146, opened 2024-12-24
Implement argument and option type casting

gustavothecoder:
As reported in the issue #129, although we have the `type` config when defining arguments and options, we don't type cast the value for the user, which for some may seem like a bug.

This PR implement a simple type casting logic for some types. Since some types that users currently receive as string will start to be received as another type (e.g. integer), this is a breaking change.

The CLI bellow demonstrates all available types:

```ruby
require 'dry/cli'

module Demo
  extend Dry::CLI::Registry

  class TypeCast < Dry::CLI::Command
    argument :a_int, type: :integer
    argument :a_flo, type: :float
    argument :a_boo, type: :boolean
    argument :a_fla, type: :flag
    argument :a_str, type: :string
    argument :a_arr, type: :array
    option :o_int, type: :integer
    option :o_flo, type: :float
    option :o_boo, type: :boolean
    option :o_fla, type: :flag
    option :o_str, type: :string
    option :o_arr, type: :array

    def call(a_int:, a_flo:, a_boo:, a_fla:, a_str:, a_arr:, o_int:, o_flo:, o_boo:, o_fla:, o_str:, o_arr:, **)
      puts <<~HEREDOC
        Arguments:
          #{a_int}\t\t\t#{a_int.class}
          #{a_flo}\t\t\t#{a_flo.class}
          #{a_boo}\t\t\t#{a_boo.class}
          #{a_fla}\t\t\t#{a_fla.class}
          #{a_str}\t#{a_str.class}
          #{a_arr}\t#{a_arr.class}
        Options:
          #{o_int}\t\t\t#{o_int.class}
          #{o_flo}\t\t\t#{o_flo.class}
          #{o_boo}\t\t\t#{o_boo.class}
          #{o_fla}\t\t\t#{o_fla.class}
          #{o_str}\t#{o_str.class}
          #{o_arr}\t#{o_arr.class}
      HEREDOC
    end
  end

  register "type-cast", TypeCast
end

Dry::CLI.new(Demo).call
```

The output:
```bash
$ demo type-cast 42 42.42 t off 'This is a string' 1 2 3 --o_int=42 --o_flo=42.42 --no-o_boo --o_fla --o_str='this is a string' --o_arr=1,2,3

# Before:
Arguments:
  42                    String
  42.42                 String
  t                     String
  off                   String
  This is a string      String
  ["1", "2", "3"]       Array
Options:
  42                    String
  42.42                 String
  false                 FalseClass
  true                  TrueClass
  this is a string      String
  ["1", "2", "3"]       Array

# After:
Arguments:
  42                    Integer
  42.42                 Float
  true                  TrueClass
  false                 FalseClass
  This is a string      String
  ["1", "2", "3"]       Array
Options:
  42                    Integer
  42.42                 Float
  false                 FalseClass
  true                  TrueClass
  this is a string      String
  ["1", "2", "3"]       Array
```

### DIFF
--- a/lib/dry/cli/option.rb
+++ b/lib/dry/cli/option.rb
@@ -70,6 +70,18 @@ module Dry
         type == :array
       end
 
+      # @since 2.0.0
+      # @api private
+      def integer?
+        type == :integer
+      end
+
+      # @since 2.0.0
+      # @api private
+      def float?
+        type == :float
+      end
+
       # @since 0.1.0
       # @api private
       def default
@@ -121,6 +133,24 @@ module Dry
           .map { |name| name.size == 1 ? "-#{name}" : "--#{name}" }
           .map { |name| boolean? || flag? ? name : "#{name} VALUE" }
       end
+
+      # @since 2.0.0
+      # @api private
+      # rubocop:disable Metrics/PerceivedComplexity
+      def type_cast(value)
+        return value if value.nil?
+
+        if integer?
+          value.to_i
+        elsif float?
+          value.to_f
+        elsif argument? && (boolean? || flag?)
+          %w[0 f false off].include?(value.downcase) ? false : true
+        else
+          value
+        end
+      end
+      # rubocop:enable Metrics/PerceivedComplexity
     end
 
     # Command line argument
--- a/lib/dry/cli/parser.rb
+++ b/lib/dry/cli/parser.rb
@@ -20,7 +20,7 @@ module Dry
         OptionParser.new do |opts|
           command.options.each do |option|
             opts.on(*option.parser_options) do |value|
-              parsed_options[option.name.to_sym] = value
+              parsed_options[option.name.to_sym] = option.type_cast(value)
             end
           end
 
@@ -77,7 +77,7 @@ module Dry
             result[cmd_arg.name] = arguments[index..]
             break
           else
-            result[cmd_arg.name] = arguments.at(index)
+            result[cmd_arg.name] = cmd_arg.type_cast(arguments.at(index))
           end
         end
 
--- a/spec/unit/dry/cli/option_spec.rb
+++ b/spec/unit/dry/cli/option_spec.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+RSpec.describe "Option" do
+  describe "type casting" do
+    it "supports :integer" do
+      o = Dry::CLI::Option.new("test", {type: :integer})
+
+      expect(o.type_cast("42")).to eq(42)
+      expect(o.type_cast("4.2")).to eq(4)
+    end
+
+    it "supports :float" do
+      o = Dry::CLI::Option.new("test", {type: :float})
+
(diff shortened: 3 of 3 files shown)


lib/dry/cli/option.rb
```diff
@@ -121,6 +133,24 @@ def alias_names
...
+      # rubocop:disable Metrics/PerceivedComplexity
+      def type_cast(value)
+        return value if value.nil?
+
+        if integer?
+          value.to_i
+        elsif float?
+          value.to_f
+        elsif argument? && (boolean? || flag?)
+          %w[0 f false off].include?(value.downcase) ? false : true
```
gustavothecoder, line 148: I used [rails](https://github.com/rails/rails/blob/main/activemodel/lib/active_model/type/boolean.rb) as a base.

noraj:
This would be really nice, because I have tons of options where I have to define booleans as string `default: 'ture', values: %w[true false]` and then call a custom `.to_bool` on the option `options[:advanced].to_bool` because user input always comes as a string.

As an alternative to having pre-defined casting, user could also provide a custom method for casting.

> Maybe something like `Dry::Types::Params` could work better here, but I'm not sure we want to pull a dependency to dry-cli.

Then could dry-types be an optional dependency of dry-cli? Needed only for using `type_cast:` on arguments and options. That way it won't pull a dependency to dry-cli by default, only for people implementing advanced use cases. Also dry-types already supports [custom types](https://dry-rb.org/gems/dry-types/1.7/custom-types/).

dry-types itself can [be extended](https://dry-rb.org/gems/dry-types/1.7/extensions/) with [maybe](https://dry-rb.org/gems/dry-types/1.7/extensions/maybe/) and [monads](https://dry-rb.org/gems/dry-types/1.7/extensions/maybe/), so how comes dry-cli couldn't be extended with dry-types? Couldn't it do `Dry::CLI.load_extensions(:types)`?

TL;DR: two options to not add default dependencies to dry-cli:

1. Support only user-defined casting, the user has to provide its own method
2. Add dry-types as an extension (so optional dependency)

katafrakt:
> dry-types itself can [be extended](https://dry-rb.org/gems/dry-types/1.7/extensions/) with [maybe](https://dry-rb.org/gems/dry-types/1.7/extensions/maybe/) and [monads](https://dry-rb.org/gems/dry-types/1.7/extensions/maybe/), so how comes dry-cli couldn't be extended with dry-types? Couldn't it do Dry::CLI.load_extensions(:types)?

The reason is that Dry CLI is kept dependency-free and the extension mechanics is coming from Dry Core.

But I like the idea of a Dry Types extension for CLI. The code is [rather small](https://github.com/dry-rb/dry-core/blob/main/lib/dry/core/extensions.rb) so we could even just pull it in and keep Dry CLI without Dry Core dependency.

katafrakt:
FYI the Dry Types implementation PR is here: https://github.com/dry-rb/dry-cli/pull/157
Feedback welcome.

katafrakt:
#157 has been merged. I'm going to close this one.

PR #148, opened 2025-01-29
Add support for Ruby 3.4

cllns:
This also drops Ruby 3.0 from CI.

The README says Ruby 2.4+ is supported. Do we really want that? I feel like dropping support for EOL'd versions of Ruby is acceptable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,10 +26,10 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
+        - "3.4"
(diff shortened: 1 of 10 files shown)


nonnenmacher:
+1 Could this be merged ?

timriley:
Yep, happy to merge this in. Thanks for getting it sorted, @cllns!

@nonnenmacher is there any reason you're +1-ing this issue? dry-cli already works on Ruby 3.4, even without this change.

PR #149, opened 2025-07-24
Add support for unlimited nested command prefixes

aaronmallen:
Previously, Dry::CLI supported grouped commands via prefixes, but only one level deep. This change extends the internal DSL to allow recursively nested `register` blocks, enabling deeply nested command hierarchies like:

```shell
my-cli foo bar baz qux
```

This improves organization for complex CLI apps with multiple levels of subcommands.

### DIFF
--- a/lib/dry/cli/registry.rb
+++ b/lib/dry/cli/registry.rb
@@ -318,9 +318,18 @@ module Dry
         # @since 0.1.0
         #
         # @see Dry::CLI::Registry#register
-        def register(name, command, aliases: [], hidden: false)
+        def register(name, command = nil, aliases: [], hidden: false, &block)
           command_name = "#{prefix} #{name}"
           registry.set(command_name, command, aliases, hidden)
+
+          if block_given?
+            prefix = self.class.new(registry, command_name, aliases, hidden)
+            if block.arity.zero?
+              prefix.instance_eval(&block)
+            else
+              yield(prefix)
+            end
+          end
         end
 
         private
--- a/spec/integration/rendering_spec.rb
+++ b/spec/integration/rendering_spec.rb
@@ -17,6 +17,7 @@ RSpec.describe "Rendering" do
         foo generate [SUBCOMMAND]
         foo greeting [RESPONSE]
         foo hello                                                  # Print a greeting
+        foo nested [SUBCOMMAND]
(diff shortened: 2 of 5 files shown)


timriley:
@aaronmallen Thanks for putting this forward! I would be very happy to have us support more deeply nested namespaces.

What's interesting about your proposal here is that it actually opens up two new features!

1. Deeply nested namespaces (per the title of this PR), but also,
2. Registering commands to the namespaces themselves

Number (2) is akin to how the `git` CLI works. `git remote` for example, executes a command that lists your remotes, while you can also use subcommands to do more specific things to your remotes, e.g. `git remote add`, etc.

This is a good power for Dry CLI to have! I want this tool to help people build the same kind of CLIs they're familiar with in the wild, and this closes an important gap.

One thing to note: with this change, we do lose the default behaviour of a namespace (itself attached to a command) automatically listing its subcommands when invoked. Previously a namespace would always print a help message with its subcommands when invoked, whereas now it may simply execute the command that's attached directly to it.

I think this is fine though, it's a necessary side effect of the new power we have, and our other built-in behaviours seem to handle this pretty well.

For example, `foo nested` shows that its subcommand may take direct arguments _or_ subcommands:

```
Commands:
  foo nested one [ARGUMENT|SUBCOMMAND]                    # A nested sub command
```

And `foo nested one --help` shows the details of both its direct command as well as its subcommands:

```
Command:
  foo nested one

Usage:
  foo nested one | foo nested one SUBCOMMAND

Description:
  A nested sub command

Subcommands:
  two                               # A nested nested sub command

Options:
  --help, -h                        # Print this help
```

If there's one thing we can add to this PR, I think it would be to enshrine this behaviour with some additional tests. This way we know we won't accidentally break the new nested subcommand and namespace-as-a-command behaviour with future changes.

Do you think you might be able to do that? After that change I'd be happy to merge!

timriley:
Actually wait, we already have tests for this! Everything to do with `root-command` and its `sub-command`, which we actually register like so:

```ruby
Foo::CLI::Commands.register "root-command", Foo::CLI::Commands::RootCommand
Foo::CLI::Commands.register "root-command sub-command", Foo::CLI::Commands::RootCommands::SubCommand
```

So in this case I think we can go ahead and merge.

timriley:
Released this in v1.3.0 :)

PR #150, opened 2025-07-24
Expose `stderr` and `stdout` for commands

aaronmallen:
Introduce `@stderr` and `@stdout` instance variables to make error and output streams accessible within commands. This enables more flexible error handling and message output customization.


lib/dry/cli.rb
```diff
@@ -97,6 +97,10 @@ def call(arguments: ARGV, out: $stdout, err: $stderr)
     # @api private
     def perform_command(arguments)
       command, args = parse(kommand, arguments, [])
+
+      command.instance_variable_set(:@stderr, err)
```
aaronmallen, line 101: The idea hear is to not have to change the signature of `Command#initialize` causing a breaking change

alassek: How about initializing a pair of `StringIO` in the command, and then

```ruby
def perform_command(arguments)
  command, args = parse(kommand, arguments, [])
  command.call(**args)
  IO.copy_stream(command.out, out)
  IO.copy_stream(command.err, err)
end
```

This would make testing the stdout and stderr of commands much simpler.

aaronmallen: The intention is to utilize the out/err passed to the CLI in the first place. dry-cil uses them to print help and errors and "inline" commands. We **should** also allow the commands to use the user specified out/err

aaronmallen:
I would even go so far as to say, it would be cool if `Dry::CLI::Banner` to the width of everything into account and aligned everything in a similar fashion.

timriley:
@aaronmallen This change looks good to me.

I'd like to merge this and then release it as v1.4.0.

Then we'll use #151 as the beginnings of 2.0, which can also include #152.

How does that sound to you? This will probably mean you need to adjust #151 so it's based off this change, rather than off the current main. Would you mind doing that? Thank you!

I'll open a convo with our maintainers just to raise awareness of this change.

timriley:
@aaronmallen Actually, one change request: can you please rename these to `@out` and `@err`?

These are the names of the args we currently expose to `Dry::CLI#call`, so it's good to be consistent. I also think it's better we don't use the literal stdout/stderr names, since these won't always be the _actual_ stdout/stderr for the current process. While it's true that `$stderr`/`$stdout` are the _default_ values, the intention is for these to be user replaceable.

timriley:
@aaronmallen Thanks for tweaking the names. FYI, I just pushed two changes:

1. Don't override `@out` and `@err` ivars if they're already defined (trying to be as cautious as possible not to break any existing usage for this 1.x version of the feature)
2. Added a test

timriley:
Will merge this now. Whatever is in the main branch from this merge we'll use as the basis for the next 1.x release. Future commits to main we can use for 2.0.

PR #151, opened 2025-07-25
Add `stderr`, `stdin`, and `stdout` support for `Dry::CLI`

aaronmallen:
Introduce `stderr`, `stdin`, and `stdout` streams to `Dry::CLI` and its components to better handle I/O. This allows commands to utilize specific input/output streams, improving flexibility and testability. Refactor existing methods to use the new stream parameters consistently.

As an alternative to #150

cc @timriley

### DIFF
--- a/lib/dry/cli.rb
+++ b/lib/dry/cli.rb
@@ -83,12 +83,13 @@ module Dry
     # Invoke the CLI
     #
     # @param arguments [Array<string>] the command line arguments (defaults to `ARGV`)
-    # @param out [IO] the standard output (defaults to `$stdout`)
-    # @param err [IO] the error output (defaults to `$stderr`)
+    # @param stderr [IO] the error output (defaults to `$stderr`)
+    # @param stdin [IO] the standard input (defaults to `$stdin`)
+    # @param stdout [IO] the standard output (defaults to `$stdout`)
     #
     # @since 0.1.0
-    def call(arguments: ARGV, out: $stdout, err: $stderr)
-      @out, @err = out, err
+    def call(arguments: ARGV, stderr: $stderr, stdin: $stdin, stdout: $stdout)
+      @stderr, @stdin, @stdout = stderr, stdin, stdout
       kommand ? perform_command(arguments) : perform_registry(arguments)
     rescue SignalException => exception
       signal_exception(exception)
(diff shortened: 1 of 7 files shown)


timriley:
Per my comments here (https://github.com/dry-rb/dry-cli/pull/150#issuecomment-3116313349), this is a change that I think is worth breaking API compatibility over, and releasing as 2.0 if needed.

Why is it breaking compatibility? Previously Dry::CLI made no claim over the structure of a command's `#initialize` method. With this change, we now expect it to receive kwargs and reserve `stdin:`, `stdout:` and `stderr:` as required keywords that we manage internally.

I think this is a good change because it sets the expectation with users that they should use the streams that are given to them, rather than handle IO themselves. This ensures we're using the same streams as given to `Dry::CLI#call`, and improves command testing, since it's clear that streams are injected, and in tests, fake IO objects can be given instead.

This arrangement matches what we already do internally within Hanami::CLI::Command, which is our own primary in-ecosystem consumer of Dry CLI.

@flash-gordon @cllns @alassek — what do you think about this? Worth the change, in your opinion? And do you actually consider it "breaking?"

lib/dry/cli.rb
```diff
@@ -145,29 +150,29 @@ def parse(command, arguments, names)
...
+      stderr.puts(result.error)
       exit(1)
     end
 
     # @since 1.1.1
     def spell_checker(result, arguments)
       spell_checker = SpellChecker.call(result, arguments)
-      err.puts spell_checker if spell_checker
+      stderr.puts spell_checker if spell_checker
       puts
```
aaronmallen, line 174: @timriley is this a bug? Should this be `stderr.puts` and if so would you like me to fix it in this PR?

timriley (commented):
This looks good to me, @aaronmallen, but it looks like we're missing some testing? These args/attributes are becoming an important part of our public API for commands, but we don't have any new tests to go along with their introduction here. Do you think you could add some in? Thanks!

katafrakt:
@aaronmallen I can add these tests if you're ok with that. Then we could merge this and, after some focus is freed up after Hanami release, perhaps we could get the 2.0 released.

aaronmallen:
@katafrakt feel free to give it a go. If not I can look at this tonight. Not sure why I never got a notification about @timriley comment

katafrakt:
Tests added, let me know if these are enough.

PR #155, opened 2026-01-27
Fix missing optional arguments to win over values

svoop:
See https://github.com/dry-rb/dry-cli/issues/126#issuecomment-3806588273

Optional arguments (either implic or with explicit `required: false`) should be accepted when missing even if a `values` list does not include `nil` as a value – at least as per the principle of least surprise. 

```ruby
argument :required_arg, required: true, values: %w(one two)
argument :optional_arg, required: false, values: %w(one two)
```

```
cmd             # ERROR
cmd one         # OKAY
cmd three       # ERROR 
cmd one two     # OKAY
cmd one three   # ERROR
```

Not sure how best to spec this, hope it's okay to having added a vacuum subcommand to the db example.

/cc @gustavothecoder

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,21 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
       # ...
     ]
   end
+  
+- As with options, optional arguments can now be omitted even if list of acceptable values is defined. (@svoop in #155)
+
+  ```ruby
+  argument :required_arg, required: true, values: %w(one two)
+  argument :optional_arg, required: false, values: %w(one two)
+  ```
+
+  ```
+  cmd             # ERROR
+  cmd one         # OKAY
+  cmd three       # ERROR 
+  cmd one two     # OKAY
+  cmd one three   # ERROR
+  ```
 
 ### Deprecated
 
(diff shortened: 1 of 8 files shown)


katafrakt:
As for spec, I was able to reproduce the bug (on main) like this: 6822f940b04aa29fbe2aecbbeceee01190894b06
Perhaps you can pull that in?

Otherwise, it looks like a great fix to me!

katafrakt:
@svoop I forgot about this PR, apologies. I pushed the test now to your branch (it needed some additional tweaking). I think we are good to go now with this?

katafrakt (approved):

katafrakt:
Let's add a changelog entry though, before merging.

svoop:
@katafrakt Thanks a bunch!

katafrakt:
Is it ready to merge? Or are we waiting for something?

svoop:
@katafrakt If the changelog is [okay like so](https://github.com/dry-rb/dry-cli/pull/155/changes/c6100f08e890d16404fc20ea47ee49069561f636), it can be squash-merged IMO.

svoop:
Thanks, @katafrakt !

PR #159, opened 2026-07-09
Add support for repeatable array options

Drowze:
Support repeatable :array options (in addition to comma-separated)

E.g.:
```ruby
# Given that `my_app` has `api` command taking a `--header` array option
# $ my_app api /some/path \
#     --header 'content-type: application-json' \
#     --header 'accept: application/json'

# Before:
options[:header] # => ["accept: application/json"]

# After:
options[:header] # => ["content-type: application-json", "accept: application/json"]
```

---

Related:
- https://github.com/dry-rb/dry-cli/pull/157#issuecomment-4925608782
- https://github.com/dry-rb/dry-cli/issues/129#issuecomment-4926140636

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 
 - Support for command namespaces (@gustavothecoder in #135)
 - New `:cast` option for options and arguments, allowing to leverage Dry Types or simple procs/lambdas to cast values from string to some other type of value (@katafrakt in #157)
+- Array options can be supplied by repeated flags. (@Drowze in #159)
+
+    The following are now equivalent:
+    ```
+    my-cli --array-flag foo,bar
+    my-cli --array-flag foo --array-flag bar
+    ```
 
 ### Changed
 
--- a/lib/dry/cli/parser.rb
+++ b/lib/dry/cli/parser.rb
@@ -22,7 +22,14 @@ module Dry
             opts.on(*option.parser_options) do |value|
               raise ValueError.new(value: value, argument: option) unless option.valid_value?(value)
 
-              parsed_options[option.name.to_sym] = option.cast(value)
+              option_name = option.name.to_sym
+              value = option.cast(value)
+              if option.array?
+                parsed_options[option_name] ||= []
+                parsed_options[option_name] += value
+              else
+                parsed_options[option_name] = value
+              end
             end
           end
 
--- a/lib/dry/cli/parser.rb
+++ b/lib/dry/cli/parser.rb
@@ -23,12 +23,11 @@ module Dry
               raise ValueError.new(value: value, argument: option) unless option.valid_value?(value)
 
               option_name = option.name.to_sym
-              value = option.cast(value)
               if option.array?
                 parsed_options[option_name] ||= []
-                parsed_options[option_name] += value
+                parsed_options[option_name] += option.cast(value)
               else
-                parsed_options[option_name] = value
+                parsed_options[option_name] = option.cast(value)
               end
             end
           end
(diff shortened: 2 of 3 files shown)


katafrakt:
I very much like the direction and I was missing ability to specify API like curl's multiple `-H` . However, I have a nit about the semantics here. This can be seen in the example you added in tests:

```
server --deps=dep42 --deps=dep43
```

I believe what would be desired in reality is to have either plural and comma-separated, or singular multiple times. So:

```
server --dep=dep42 --dep=dep43
# or
server --deps=dep42,dep43
```

This makes me think they are not, in fact, interchangeable. Maybe instead of reusing `:array` type we should have `:multiple` flag/type here?

Please, don't take it as a change request straight away. I'd like it to be discussed first.

Drowze:
Hey @katafrakt, thanks a lot for the thoughtful feedback! 😄 

My intention indeed was to mimic other common CLI tools - so definitely no coincidence that this is very similar to `curl`. Glad you've noted and liked this direction as well!

> believe what would be desired in reality is to have either plural and comma-separated, or singular multiple times

I can see where you're coming from... Indeed the example with repeatable pluralized options is a bit unfortunate.

Also, not sure if that's what you suggested, but having an _actual_ decent pluralization built-into array types would probably be tricky, so my vote would be against that direction (and options already support `aliases` btw, so I don't see a good reason to have inflection support built into dry-cli).

> This makes me think they are not, in fact, interchangeable. Maybe instead of reusing :array type we should have :multiple flag/type here?

Funny you mentioned that :) my initial implementation was using a new `:repeatable` flag. The main reason me to change the direction and re-use the `:array` type is that I wasn't sure _how_ to display the description of repeatable options (in the "command banner" e.g. when calling a command with `--help`).

For context: I didn't want to inadvertently make the description too big (especially since `dry-cli` does not have any "auto-wrapping" on the banner) so, with that in mind, my solution in that original branch was to simply append a ` (repeatable)` to the description of those options.
I was a bit on the fence about how it turned out though, which is why in the end I've decided to open this pull request simply reusing the `:array` type (to avoid worrying about descriptions 😅 )

Here's my original solution using a `:repeatable` flag: https://github.com/dry-rb/dry-cli/compare/main...Drowze:dry-cli:rg-support-repeatable-options
Please let me know your thoughts on it! I have no objections on closing this PR and re-opening from that branch if you prefer to :+1:

timriley:
This is a good feature, thank you @Drowze! 

I think that using `:array` type here feels like the most natural approach, even if the option name isn't the most perfect fit for the single-option-comma-separated or the repeated-option inputs.

I see type for an option being about the _value_ it's meant to hold, and both of these input styles lead to arrays.

In looking into this, I found a related bug that I'm going to fix independently via another PR, and then I'll rebase this and make sure it's ready to merge. Thanks for bearing with me :)

Drowze:
Thanks for the feedback @timriley 🙌 

Just rebased the PR as there was some rubocop infractions which were fixed on latest `main` :)

timriley:
Thanks! Just for the record, I've pushed #161 and #162, which both deal with various bugs in our array handling. Next step will be to merge these and then ensure this one still applies smoothly over the top.

timriley:
Alright, I've rebased this over main now that I've merged those other PRs. As part of resolving the conflicts, I decided to keep all the OptionParser code together, rather than extracting a method, since that means it's only one place to see everything we're giving to OptionParser.

I've pushed this up to this PR's branch. It's a nice small change now. @Drowze — does this all look OK to you?

Drowze:
@timriley Hey it's looking great to me :) no objections at all. Creating a new method for `OptionParser` was actually motivated by some RuboCop Metrics infractions tbh - no objections on keeping it out.

timriley:
Thanks @Drowze! ❤️

PR #160, opened 2026-07-31
Add `long_desc` and split `-h` from `--help`

aaronmallen:
Commands and namespaces can now set a long description with `long_desc`. `-h` prints the
short description from `desc`; `--help` prints the long one. This mirrors how Clap treats
`about` and `long_about` in Rust.

```ruby
class Echo < Dry::CLI::Command
  desc "Prints given input"
  long_desc <<~DESC
    Prints the given input back to standard output.

    When no input is given, it prints an empty line.
  DESC
end
```

```sh
$ foo echo -h shows "Prints given input". $ foo echo --help shows the full text.
```

## How

- Command and Namespace gain a long_desc class method that stores @long_description.
- Parser registers -h and --help as separate handlers and marks the result when the
user asked for long help.
- Banner picks the description by help mode: long help uses long_description and falls
back to description; short help always uses description. Multi-line text now indents
under the Description: heading, and blank lines stay blank.
- Command listings still use the short desc.

## Compatibility

Commands without long_desc print the same output for both flags, so nothing changes for
existing CLIs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Break Versioning](https://www.taoensso.com/break-ve
 
 - Support for command namespaces (@gustavothecoder in #135)
 - New `:cast` option for options and arguments, allowing to leverage Dry Types or simple procs/lambdas to cast values from string to some other type of value (@katafrakt in #157)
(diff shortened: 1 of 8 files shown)


timriley:
Thanks @aaronmallen!

PR #165, opened 2026-08-15
Allow command options to be added externally

### LINKED ISSUE #164: Allow command options to be added externally
In hanami-cli we have generator commands like `generate action`.

Then in hanami-rspec and hanami-minitest, we add hooks to those commands, like this:

```
Hanami::CLI.after "generate action", Commands::Generate::Action
```

Those hooks want a specific extra option for them to do their job, in this case `--skip-tests`.

Right now the only way to support this extra option is to declare it on the _base_ command in hanami-cli, even though nothing inside that project needs that option.

It would be nice if we could find a way for extra options to be added externally, so projects like hanami-rspec can add them specifically for their own extensions, and the base command can remain unchanged.

afomera:
> [!NOTE]
> Per the [Hanakai AI Policy](https://hanakai.org/ai-policy): I used AI to help me put together the code changes with my heavy involvement and direction, but I have reviewed every line of code until it reached the quality bar I set for myself for this work.
> 
> At no point was AI running without my direction or supervision.
> 
> I believe the comments I've left are useful and the specs feel like they've ensured things are tested properly to me, but it is a lot of them.  The commit message + PR description was written by me.

In hanami-cli, `generate action` declares a `--skip-tests` flag that nothing in hanami-cli itself uses. It exists only so that hanami-rspec and hanami-minitest, which hook into the command via `Hanami::CLI.after "generate action", ...`, have a flag their hooks can read. An unparsed flag is a hard error, so there was no other way for those gems to accept one.

Add `Registry#command`, which returns the registered command, along with `Registry#option` and `Registry#argument` as guarded wrappers over it. A third-party gem can now contribute the param its own hook needs

```ruby
Hanami::CLI.after  "generate action", Commands::Generate::Action
Hanami::CLI.option "generate action", :skip_tests,
                    type: :flag, default: false, desc: "Skip test generation"
```

Because more than one gem may want the same option, adding it twice is a no-op rather than an error, as long as both declarations agree on :type, :required, :values and :default. Otherwise IncompatibleOptionError names the settings that differ. :cast is not compared, because procs aren't meaningfully comparable.

Note that this extends the command class, so it is not scoped to a single registration, and a subclass defined before the call won't inherit the param.

Registry already had a private #command returning a LookupResult, **which is renamed to #lookup.**

Commands and callbacks are now also passed only the params their #call declares, so neither side needs a ** catch-all to tolerate params contributed by another gem. Params are passed through untouched unless #call declares keywords and no keyword splat, so this only affects signatures that raise ArgumentError today: a #call declaring no keywords at all (def call(*args), or a callback block) still receives every param, since it reads them from a single positional Hash.

Closes #164


timriley:
Thanks so much for this, @afomera! I fixed a few edge cases and did a bit of refactoring (see commits above). Merging this in now! Also happy to take feedback and adjust post-merge if you have any thoughts on the changes :)

timriley:
Also, would you be up for changing hanami-rspec and hanami-minitest to use this new API? And getting rid of the unneeded splat from the commands in hanami-cli?

PR #166, opened 2026-08-27
Add text styles with true color, degraded to fit the terminal

aaronmallen:
## A style is a value

Chain style names to build one, give it a name, then apply it as often as you like. Chaining returns a new style each time, so a style is safe to share and reuse.

```ruby
ERROR = Dry::CLI::Style.bold.red
MUTED = Dry::CLI::Style.dim

ERROR.call("Boom")     # => #<Dry::CLI::Style::Text(bold.red) "Boom">
ERROR["Boom"]          # same
ERROR["Boom"].to_s     # "\e[1;31mBoom\e[0m"
names.map(&MUTED)      # styles are procs
```

Pass the text straight to a style name for one-off use, and nest styles freely: an outer style reopens itself after an inner one resets.

```ruby
Dry::CLI::Style.bold.red("Boom")
Dry::CLI::Style.bold.call("hello #{ERROR.call("world")} again").to_s
# => "\e[1mhello \e[31mworld\e[0m\e[1m again\e[0m"
```

## Styled text only renders when sent to a stream

Styled text is returned to you as a `Dry::CLI::Style::Text`. Only when that text is sent to a stream (now represented as a `Dry::CLI::Stream`) is it rendered, and only based on that stream's color support.

For example, for a CLI invocation of `mycli 2> errors.txt`, then text sent to `stdout` will be styled (since it is a terminal), but text sent to `stderr` is _not_ (since it is a file).

## Colors degrade gracefully to the terminal capabilities

Write the color you want. We work out what the terminal can show and swap in the closest match, so one constant is right everywhere.

```ruby
ERROR = Dry::CLI::Style.bold.rgb(255, 0, 0)

ERROR.call("Boom")
# 24-bit   => "\e[1;38;2;255;0;0mBoom\e[0m"
# 256      => "\e[1;38;5;196mBoom\e[0m"
# 16       => "\e[1;91mBoom\e[0m"
# 8        => "\e[1;31mBoom\e[0m"
# no color => "\e[1mBoom\e[0m"
```

Colors come as `rgb`, `hex` (3 or 6 digit), `ansi256`, or by name: the eight ANSI colors and their `bright_` versions. Each has an `on_` variant for the background. Bad input raises `InvalidColorError`.

## Turning it off

Styling is on when the program writes to a terminal and `NO_COLOR` is unset. `Dry::CLI::Style.enabled = false` forces it off, and when it is off a style returns its text with any styling stripped, so output stays clean.

`Dry::CLI::Style.color_level` reports what the terminal can show and takes an override, for a `--color` flag of your own or to pin output in tests. `FORCE_COLOR` of `0` to `3` names a level too.

## In commands

`Dry::CLI::Command` gets `style` and `unstyle` at both class and instance level, so a command can declare its styles up front:

```ruby
module MyCLI
  module Styles
    extend Dry::CLI::StyleMixin

    ERROR = style.bold.hex("#cc0000")
    MUTED = style.dim
  end
end
```


alassek (approved):
🥇 This is great, precisely the sort of API I would want.

lib/dry/cli/style.rb
```diff
@@ -171,17 +171,6 @@ def color_level
...
 
-        # @!method colour_level
-        # @!method colour_level=(level)
-        #
-        #   British spelling of {color_level} and {color_level=}. The two names are the same
-        #   method, so either reads what the other set.
-        #
-        #   @api public
-        #   @since x.y.z
-        alias_method :colour_level, :color_level
```
aaronmallen: I mean I understand we generally follow US spelling... but we lose nothing by leaving these in.

timriley:
Thanks @aaronmallen for moving this forward! And thanks @alassek for the first look, I'm glad you've liked it so far :)

Could I please ask you to have another check? I've pushed up a few refactorings, as well a significant functional change — now a string sent through one of the styles comes back as a `Dry::CLI::Style::Text`, and the styles are only rendered when we sent that text to a stream (which we now wrap as a `Dry::CLI::Stream`). This moves the "do we style this text?" question from being global to stream-based, which is important for when a CLI program is sending one stream to the terminal but another to a file. In this case, we want the terminal output to be styled, but the file contents to be unstyled. [0b58067](https://github.com/dry-rb/dry-cli/pull/166/commits/0b5806715040dc4b61421a5a39da8182150ca32a) and [4fbc8c4](https://github.com/dry-rb/dry-cli/pull/166/commits/4fbc8c40f0f4cf63769095643c27a04064de50ae) are the commits taking care of this change.

spec/support/styling.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# Ensure that any existing style-related env vars do not conflict with our styling tests.
+RSpec.configure do |config|
+  config.before :suite do
+    ENV.delete("NO_COLOR")
+    ENV.delete("FORCE_COLOR")
```
katafrakt, line 7: Should we use something like ClimateControl (in the future, not necessarily in this PR) to manage ENV in tests?

katafrakt (approved):
Looks like a great addition to Dry::CLI!

timriley:
I checked in with @aaronmallen about the progress here, and he asked for one thing clarified:

> My only concern (and I'm unsure if your code violates this concern) is that we should not spend compute on encoding a string when NO_COLOR is set. If I'm reading this correctly you are retroactively striping styles from the string when it is determined they shouldn't be applied.

So the key change is that sending a string through the styling methods now returns a `Dry::CLI::Style::Text` instance. Styling doesn't happen immediately; it's deferred until we output that text to a stream (stdout or stderr). This is important to fulfil a behaviour that wasn't present in my original prototype: when a CLI is called like `mycli 2> errors.log`, then stdout should be styled (it's going to the terminal), but stderr should be _unstyled_ (it's going to a file). In this arrangement, styles are _only_ added when they're needed. In the case of a stream that is not a terminal, the style codes are never added in the first place when rendering the text, so no stripping is required.

As for the `NO_COLOR` case (styling globally disabled), these `Dry::CLI::Style::Text` instances are still created, but they all short-circuit any style-related logic:

- `Dry::CLI::Stream` knows its `#color_level` via `Dry::CLI::Style.level_for` (which is `nil` when `NO_COLOR` is set)
- When the stream is outputting text, it passes this level to `Dry::CLI::Style::Text#render`
- `Dry::CLI::Style::Text#render` has a fast early exit when given a level of `nil` — it returns `#plain`, which just directly gives back the text on its own, without any styling

So we do everything we can to avoid any real work in this case. The only cost having the text live as this `Style::Text` object until the moment of rendering, but I think that is a worthwhile tradeoff for the features it gains us (not just the ability to style based on each stream's capabilities, but also things like natural concatenation of styled/unstyled text, e.g. `style.bold.red["ERROR: "] + the_error_text`).

Does that satisfy your concern, @aaronmallen?

aaronmallen:
@timriley :shipit:

afomera (approved):
love it!

PR #167, opened 2026-09-03
Free up `#initialize` for command subclasses

timriley:
Free up `#initialize` in command subclasses so they can use it for their own purposes only.

`stdin:`, `stdout:`, and `stderr:` are still injectable dependencies, but these are handled by `.new`, which sets them as instance variables directly, before forwarding all other args onto `#initialize`.

Aside from reducing boilerplate for users in their command subclasses, this also decreases the risk of bugs. Without this, every command subclass that adds its own `#initialize` needs to remember to accept `**args` and then call `super(**args)`, otherwise the streams would never be set.

I thought of this change as part of updating hanami-cli to use the new streams provided as part of dry-cli v2. Now that I've made this change here, my next step will be to show what the updated hanami-cli commands look like. Stand by for that, but in the meantime, I'm still interested in feedback on this change in isolation :)

### DIFF
--- a/lib/dry/cli/command.rb
+++ b/lib/dry/cli/command.rb
@@ -9,6 +9,29 @@ module Dry
   class CLI
     # Base class for commands
     #
+    # ## Streams
+    #
+    # A command should write to the streams it is given, available as {#stdout}, {#stderr} and
+    # {#stdin}. These are set before `#initialize` runs, so you can access them in your
+    # `#initialize` as required.
+    #
+    # A command registered as an instance (instead of a class) is built before the CLI knows where
+    # its output should go, so at that point {#stdout} and {#stderr} fall back to Ruby's standard
+    # `$stdout` and `$stderr`. The CLI's real streams arrive later, when the command is called, so
+    # you should build on top of streams only when you use them:
+    #
+    # ```
+    # # Wrong: built at initialization; captures the fallback stream
+    # def initialize
+    #   @logger = Logger.new(stdout)
+    # end
+    #
+    # # Right: built on first use; captures the stream the CLI is using
+    # def logger
+    #   @logger ||= Logger.new(stdout)
+    # end
+    # ```
+    #
     # @since 0.1.0
     class Command
       include StyleMixin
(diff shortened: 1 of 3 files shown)


parndt (approved):
I like this!

aaronmallen (approved):
This is good soup

afomera (approved):

timriley:
Thanks for checking this out, folks! As it turns out, I discovered an additional need while continuing the hanami-cli updates: making it it possible for users of Dry CLI to specify their _own_ "auto-assigned" keywords like we have here with `stdout:`, etc.

To support this, I've introduced an "auto-assigned keywords" feature. See it in the last commit above. The lowdown:

This feature formalises the behaviour we introduced where the `stderr:`, `stdin:`, `stdout:` args were assigned to ivars without subclass `#initialize` methods needing to worry about receiving those as `**args` and forwarding them to `super`.

Now our users can take advantage of this behavior. It’s common for a CLI app to require some “standard” dependencies be provided to every command class through their `#initialize` methods. These can now be set up like so:

```ruby
# In a CLI app’s base command class

# `inflector:` is a standard dependency
def self.auto_assign_keywords = super + %i[inflector]

# Supply default values here
private def auto_assign(inflector: Dry::Inflector.new, **kwargs)
  super(**kwargs)
  @inflector = inflector
end
```

The above will make `inflector:` work as an argument to `.new`, and assign its value to an ivar automatically (via `#auto_assign`) before `#initialize` is called.

---

We're using this for the `fs:` dependency in hanami-cli. See `lib/hanami/cli/command.rb` in https://github.com/hanami/hanami-cli/pull/447, where we do this:

```ruby
def self.auto_assign_keywords = super + %i[fs]

# ...

private

def auto_assign(fs: nil, **kwargs)
  super(**kwargs)
  @fs = fs
end
```

This means we don't have to worry about adding `**args` and `super(**args)` to every single command subclass in Hanami CLI.

This is an example of Dry CLI working as "a framework for a framework", allowing CLI apps to implement their own common baseline behaviour without having to foist any boilerplate on their _own_ internal command classes.

lib/dry/cli/command.rb
```diff
@@ -439,24 +462,58 @@ def self.superclass_options
...
+      # A subclass adding its own should add to this list and assign them in {#auto_assign}:
+      #
+      # ```
+      # def self.auto_assign_keywords = super + %i[inflector]
+      #
+      # private def auto_assign(inflector: Dry::Inflector.new, **kwargs)
+      #   super(**kwargs)
+      #   @inflector = inflector
+      # end
+      # ```
```
alassek: Since this would be trivial enough to metaprogram `auto_assign`, is your intention here to make it difficult on purpose to discourage overuse?

alassek (approved):
I have felt the awkwardness of this problem myself, and I think this is a reasonable approach.

lib/dry/cli/command.rb
timriley: Glad you like the feature, @alassek! And to answer your question: yes, my initial thinking was (a) keep the implementation simple, and (b) add a bit of friction as part of ensuring it isn't over-used.

But your question made me rethink things, both the implementation as well as name. I've just pushed up some adjustments:

- Renamed the key method from `#auto_assign` to `#auto_initialize`, to make it clearer at which point in the object lifecycle it is called. I also added docs to encourage class authors to place it right alongside their `#initialize` method.
- Switched to inferring the `#auto_initialize` args by inspecting params and walking up the chain of super methods. In practice, there won't ever be too much work to do here, and it does make the whole job simpler for our users.

If you're still happy with these changes, I'll merge this in and get back onto updating Hanami CLI to use all the new Dry CLI features :)

timriley:
FYI, I pushed up an adjustment to how this works — see https://github.com/dry-rb/dry-cli/pull/167#discussion_r3986141632.

timriley:
I'm going to go ahead and merge this one now — we're in agreement about this being a useful facility, and I'm happy with the new naming and inference of parameters. Happy to take feedback post-merge and adjust things further, too!

lib/dry/cli/command.rb
alassek: Sorry I am late getting back to this, but yes, I think the new approach looks good too. I think the complexity of inspecting params is warranted.

PR #168, opened 2026-09-22
Dry::CLI::Style::Text formatting support

alassek:
This implements `Text#%` that mirrors the behavior of `String#%` while respecting part boundaries.

AI Disclosure: I wrote the implementation by hand, but utilized a DeepSeek agent to identify edge-cases and assist building out tests.


lib/dry/cli/style/formatter.rb
```diff
@@ -0,0 +1,110 @@
...
+      # while respecting part boundaries.
+      #
+      # @api private
+      module Formatter
+        # Formatting re-applies the full argument list while walking each part, so
+        # Ruby warns about this file whenever a part consumes fewer arguments than
+        # are supplied. Filter out that one warning rather than mutating $VERBOSE,
+        # which is global and cannot be made thread-safe. This intentionally
+        # bypasses downstream Warning handlers for this message.
+        ::Warning.singleton_class.prepend(Module.new do
```
alassek, line 16: The `warning` gem is being removed from the standard lib to rubygems, so we can't assume its presence. This pattern was adopted from [dry-monads](https://github.com/dry-rb/dry-monads/blob/3711a198516bf22976e565aaef646303e22a5317/lib/dry/monads/do/all.rb#L158)

lib/dry/cli/style/formatter.rb
```diff
@@ -0,0 +1,110 @@
...
+
+          # Incremental string format, with an internal cursor.
+          #
+          # @param text [#to_s] Text part that may contain formatting directives
+          # @param args [Object] formatting arguments
+          #
+          # @return [String] the formatted part
+          def apply(text, args)
+            self << text.to_s
+            args = [args] if args.is_a?(Hash) && !named_references?
```
alassek: This is a subtlety of MRI vs JRuby but the extra checking passes both.

aaronmallen (approved):
:shipit:

lib/dry/cli/style/formatter.rb
timriley: @alassek Thank you for taking this into account!

I'm happy with the code as it is, but I do have a couple questions:

Do you have a reference somewhere that tells of the plans to move `Warning` out of stdlib? I couldn't find anything. I'm aware there is a "warning" gem in RubyGems, but that is a third-party gem (from Jeremy Evans) that provides additional ways to customise warnings. Given that `Warning.warn` is called from `Kernel#warn`, I'd be very surprised if it ended up going away.

And if it is true that it's moving out of stdlib, shouldn't we have some code here that conditionally checks for the `Warning` constant having been defined?

timriley: Feels like the pattern that we're following from Dry Monads is the same one advertised [in the `Warning` module docs](https://docs.ruby-lang.org/en/4.0/Warning.html):

> If you want to change the behavior of [Warning.warn](https://docs.ruby-lang.org/en/4.0/Warning.html#method-i-warn) you should use Warning.extend(MyNewModuleWithWarnMethod) and you can use super to get the default behavior of printing the warning to $stderr.
> 
> Example:
> 
> ```ruby
> module MyWarningFilter
>   def warn(message, category: nil, **kwargs)
>     if /some warning I want to ignore/.match?(message)
>       # ignore
>     else
>       super
>     end
>   end
> end
> Warning.extend MyWarningFilter
> ```

I haven't looked into it why we did it using `singleton_class.prepend` in Dry Monads vs just plain `.extend`, but I presume it's because we really wanted to make sure we intercepted the call first.

lib/dry/cli/style/formatter.rb
```diff
@@ -0,0 +1,110 @@
...
+          end
+        end)
+
+        class << self
+          # Apply formatting to a Style::Text object.
+          #
+          # Arguments are normalized the way String#% sees them: an object that
+          # converts to an array is applied as positional arguments, an object
+          # that converts to a hash is applied as named arguments, and anything
+          # else is applied as a single value.
```
timriley, line 34: This reads really nicely, kudos!

lib/dry/cli/style/text.rb
```diff
@@ -116,6 +116,22 @@ def +(other)
           other.is_a?(Text) ? Text.new(parts + other.parts) : Text.new(parts + [[nil, other]])
         end
 
+        # Apply arguments as a formatting string.
+        #
+        # @see String#%
+        #
+        # @example
+        #   format = dim["Waiting for "] + bold["%d"] + dim[" seconds"]
+        #   format % 5
```
timriley: This could be an even more helpful example if it shows the return value too?

lib/dry/cli/style/formatter.rb
timriley: Worth adding a comment to this effect above it?

spec/unit/dry/cli/style/formatter_spec.rb
```diff
@@ -0,0 +1,244 @@
...
+  def format_text(input, args)
+    described_class.call(input, args)
+  end
+
+  describe "array arguments" do
+    it "applies one argument list across parts in order" do
+      styled = red["a %s"] + " and %s"
+      result = format_text(styled, ["x", "y"])
+
+      expect(result.parts).to eq [[red, "a x"], [nil, " and y"]]
```
timriley, line 25: Being able to inspect the parts and still use the style objects as part of the testing is really quite nice!

timriley (approved):
Brilliant work! Couple minor thoughts for you. Happy for you to merge once you've considered them :)

lib/dry/cli/style/formatter.rb
timriley: I think I realise the possible confusion here: `Warning.warn` has existed as a customisation point [since Ruby 2.5](https://idiosyncratic-ruby.com/65-warning-the-module.html). The [warning gem](https://rubygems.org/gems/warning), OTOH, is just a set of optional enhancements over Ruby's built-in warnings (like allowing selective silencing via a convenient API), but it's never been bundled with Ruby.

So in our gems, we can't ever use the warning gem. Instead, we use the recommended way of customising Ruby's built-in `Warning.warn`, which is what we're doing here.

So all of this is to say that I think the code you have here is good, that no changes are needed, and that I don't have any further questions to ask :)

alassek: I could've sworn I was seeing deprecation warning about this, but I forget where now. Must've confused different things.

alassek:
@timriley significant refactor to the JRuby normalization, worth another look. Does that address everything?

timriley:
@alassek Love what you did with that one. The method and explanation make everything super clear and will help our future selves make better decisions when maintaining this. ❤️ Plus we get the bonus of skipping out on the work if we're not on JRuby.

This looks good to merge!

timriley:
(I wonder if we should also file a bug against JRuby about this?)

